PR #4, opened 2020-03-11
Refactor the card pattern to utilise Vanilla more

anthonydillon:
## Done
Removed the custom card styling and replaced it with the native Vanilla card with the addition of two utilities which have to propose to Vanilla.

@solazio I wanted to show how I would do it so a PR seemed the best way. Please, pick it apart. 

## QA
- See that the cards match [the design](https://app.zeplin.io/project/5d3f1850b938ee0dc24c60c4/screen/5e3984c0bdbcb10d1946c78f)


solazio:
@anthonydillon thanks. Definitely a better solution than mine. My only comment is that the images have to be round, so I would need to add some custom css to the card pattern. Or you have any other suggestion for that?

anthonydillon:
I do, I would [extend the media objects style](https://vanillaframework.io/docs/patterns/media-object#circular) and create a global `.is-round` class as it looks like avatars are likely to be all circular.

I would then propose a missing avatar pattern in vanilla.

PR #5, opened 2020-03-12
Add side navigation basic functionality

### LINKED ISSUE #3: Add side navigation functionality

solazio:
## Done

- Add basic functionality to the side navigation

## QA

- Compare to [design](https://app.zeplin.io/project/5d3f1850b938ee0dc24c60c4/screen/5e39844ac39722c6d40e682a) - note that I used the future vanilla side navigation
- Test it with javascript disabled as well - here is a [guide to disable it](https://developers.google.com/web/tools/chrome-devtools/javascript/disable)

## Issue / Card

Fixes #3


static/js/side-navigation.js
```diff
@@ -0,0 +1,33 @@
+(function() {
```
Lukewh: Is this JS needed? I think we should make everything work without JS first as we need to figure out how we want to structure and make the JS maintainable with tests.

static/sass/styles.scss
```diff
@@ -46,9 +46,10 @@ $theme-default-nav: dark;
 @include p-charmhub-strip;
 @import 'charmhub_footer';
 @include charmhub-p-footer;
-@import 'charmhub_sidenav';
-@include charmhub-p-sidenav;
+@import 'patterns_side-navigation';
```
Lukewh: Maybe add a note that this can be removed when sidebar nav lands in vanilla?

Lukewh:
Clicking on one of the categories causes a page load, clicking the x to remove it doesn't - again I think we need to focus on the non-js version first.

webapp/helpers.py
```diff
@@ -17,12 +17,18 @@ def join_filters(filters):
...
-        filter_string.append(f"{filter_type}={filter_type_filters}")
+        if len(filters[filter_type]) > 0:
+            filter_type_filters = ",".join(filters[filter_type])
+            filter_string.append(f"{filter_type}={filter_type_filters}")
 
-    filter_string = "&".join(filter_string)
+    if len(filter_string) == 0:
+        filter_string = "/"
+    elif len(filter_string) == 1:
+        filter_string = f"?{''.join(filter_string)}"
```
jkfran: If it is just one there is no need to use join, right?

clagom:
If you scroll down and click on a filter from the left nav then the page jumps on top. Is there a way we can avoid that?

solazio:
> If you scroll down and click on a filter from the left nav then the page jumps on top. Is there a way we can avoid that?

As far as i know, you cannot avoid that when if the page reloads... we could do it using JS instead. @Lukewh what do you think?

Lukewh:
> If you scroll down and click on a filter from the left nav then the page jumps to the top. Is there a way we can avoid that?

@clagom when we implement the JS later on, this won't happen. Currently clicking reloads the page (they're just regular links)

clagom:
Sounds good

PR #6, opened 2020-03-12
Add husky and lint-staged

solazio:
## Done

- Add husky and lint-staged

## QA

- Run it locally, make some changes and then commit. Note husky does the job and runs the linters

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,20 @@
     "test": "yarn run lint-scss && yarn run lint-python && yarn run test-python",
     "test-python": "python3 -m unittest discover tests"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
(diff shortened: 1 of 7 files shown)


hatched:
fwiw we've been removing husky from our other projects because we find that forcing people to lint every commit was not very welcome to everyone's workflow. Those who wanted it typically already have it as part of their editors.

solazio:
thanks for letting us know @hatched. In that case @Lukewh shall I revert the change?

Lukewh:
@hatched which projects have we removed husky from? Was there a conversation about it that I missed? I'm keen to keep it on charmhub and snapcraft - I think it's really valuable.

anthonydillon:
I believe it was agreed its a per-project decision. In this case, I think it's useful as the websites use it.

PR #10, opened 2020-03-16
1270 details page header

solazio:
## Done

- Add `/details` page header

## QA

- View the site locally in your web browser at: http://0.0.0.0:8045/charm
- Compare to [design](https://app.zeplin.io/project/5d3f1850b938ee0dc24c60c4/screen/5e2abd08b530e498135c80b6)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1270
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1263


.eslintrc.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
hatched: Do we have a common shared eslint config somewhere?

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
+function initTabs() {
+  var tabLinks = document.querySelectorAll("a[href^='#']");
+
+  // prevent the page jumping around when switching tabs,
+  // whilst still using :target
+  // https://gist.github.com/pimterry/260841c2104f27cadc954a29b9873b96#file-disable-link-jump-with-workaround-js
+  [].forEach.call(tabLinks, function(link) {
```
hatched: a `NodeList` now has a `forEach` method, so using this trick is no longer necessary.

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
...
+    function() {
+      setActiveTab();
+    },
+    false
+  );
+
+  function setActiveTab() {
+    var urlHash = window.location.hash;
+    [].forEach.call(tabLinks, function(link) {
+      if (urlHash) {
```
hatched: This conditional can be moved up a level, no need to loop through each link if there is no hash.

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
...
+    "hashchange",
+    function() {
+      setActiveTab();
+    },
+    false
+  );
+
+  function setActiveTab() {
+    var urlHash = window.location.hash;
+    [].forEach.call(tabLinks, function(link) {
```
hatched: See above about `forEach`

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
...
+      if (urlHash) {
+        if ("#" + link.getAttribute("aria-controls") === urlHash) {
```
hatched: Using this approach for a moment you may have multiple links with `aria-selected === true`. You might want to start with setting all of them to false first then using `querySelector` to grab just the one that matches the `urlHash`

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
...
+
+  window.addEventListener(
+    "hashchange",
+    function() {
+      setActiveTab();
+    },
+    false
+  );
+
+  function setActiveTab() {
```
hatched: as a personal preference I like to pre-hoist the functions so you don't have to scroll through the file looking for content that's not in a function.

.eslintrc.js
solazio: @hatched we don't have, at least as far as I know. Perhaps we should create one. Will add this topic to the agenda for the next Dev catch-up.

static/js/tabs.js
```diff
@@ -0,0 +1,44 @@
+function initTabs() {
```
Lukewh: Hmm, I guess tabs is where no-js falls down.
I'd like to find a way to get this to work without JS - as with the homepage. Perhaps we can have the tabs load the page again, with sub-pages like:
`/charm/` -> Overview
`/charm/configuration` -> Configuration
etc.
Then when we add the JS, we can intercept these loads.

anthonydillon:
In the pass I have displayed all the content and the tabs should work by jumping to the sections via inpage anchors. If js kicks in it hides all sections selects the first tab init the tab functions.

We should update the Vanilla example to include this.

clagom:
- Overview should be the default opened tab, now I need to click on 'overview' but should be displayed as the first thing.
- I know the design implementation is still WIP, but the code snippet (juju deploy openstack-dashboard) in the header shouldn't have any background colour

spencerbygraves:
Thanks @solazio, looking good.

I think the vertical spacing would be better if we applied `u-no-margin--bottom` in two places please:

image

image

Also, as @clagom says, the code snippet shouldn't have a background colour.

Thanks.

hatched:
Ahh I wasn't aware we wanted this to work without JS. It's entirely possible to make a tabbed layout with only CSS so maybe you could do that instead then enhance further (if wanted) if JS is loaded.

clagom:
Can we restart the demo please

PR #13, opened 2020-03-18
Add actions tab on the details page

solazio:
## Done

Add actions tab on the details page

## QA

- View the site locally in your web browser at: http://0.0.0.0:8045/charm#actions
- Compare to [design](https://app.zeplin.io/project/5d3f1850b938ee0dc24c60c4/screen/5e2abd087eecc454cf8abf82)
- There is no js for the accordion to work yet. This will be added once we implement actions for all the accordion buttons

## Issue / Card

Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1265

### DIFF
--- a/static/sass/_pattern_p-accordion.scss
+++ b/static/sass/_pattern_p-accordion.scss
@@ -11,7 +11,6 @@
         font-size: 0.875rem;
       }
     }
-
     .p-accordion__panel {
       padding-left: 0;
 
@@ -28,7 +27,6 @@
         &--1 {
           padding-left: $spv-inner--large;
         }
-
         &--2 {
           padding-left: $spv-inner--large * 2;
         }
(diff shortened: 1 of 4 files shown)


Lukewh:
Could you copy and paste the content from the open accordion to the others, so we can see what it's like without JS? Thanks

clagom:
> Could you copy and paste the content from the open accordion to the others, so we can see what it's like without JS? Thanks

Will we still have the + and - accordion icons even if without JS they don't work?

Lukewh:
Without JS the icons still show - can we hide them?

spencerbygraves:
>Without JS the icons still show - can we hide them? We should also remove the interactive feel (hover, focus etc).

Let's see how it looks. I don't think we considered a non JS version of the accordion, the sections may be less obvious without the icons. Especially as we have h5 after the titles.

Lukewh:
Let's just roll with it like this for now. We're going to have to revisit some stuff anyway.

PR #20, opened 2020-03-23
Convert to floats instead of flex

anthonydillon:
## Done
Lose the u-flex stuff for floats

## QA
- Check it all looks ok

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -38,6 +38,7 @@ $theme-default-nav: dark;
 
 // Vanilla utilities
 @include vf-u-align;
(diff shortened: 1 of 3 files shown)


static/sass/styles.scss
```diff
@@ -48,10 +49,6 @@ $theme-default-nav: dark;
 @include vf-u-vertically-center;
 
 // Import site specific patterns and overrides
-@import "pattern_p-card";
-
-@include p-charmhub-card;
```
anthonydillon: Removed this as it will need a better solution when applying the platform icons.

PR #35, opened 2020-03-30
Upgrade vanilla to 2.9.0 & add side navigation to home page

### LINKED ISSUE #31: Upgrade vanilla version & update side navigation on the home page

solazio:
## Done

- Upgrade to vanilla 2.9.0
- Add vanilla side-navigation to the homepage

## QA

- See the side navigation works and looks as expected

## Issue / Card

Fixes #31 

## Note

@spencerbygraves  as there are no such colours for the side navigation in vanilla, do we want to stick to the default colours, or we want to implement charmhub colours in vanilla?


solazio:
>  Code looks good. Will any of the custom CSS you've added be upstreamed to Vanilla at any point? Might be worth adding a note in the code that it's something to keep track of?

not sure how Spencer wants to do it, but I will add a note if we decide to add it to vanilla.

spencerbygraves:
@solazio @Lukewh as the side navigation is a new component, it would be good to use it here without overrides for colours.

The light theme doesn't yet contain opacity for the hover and selected states, so I don't think it will look good on the background colour we have now.

Can we change the background to white please, and use the side nav as intended. This can be something we come back to if needed.

We can propose the toggle on selected items to Vanilla later on.

This is how it will look after the update:

image

Thanks.

PR #39, opened 2020-03-31
Add dummy data to python

### LINKED ISSUE #36: Add dummy data to python and pass through to the template

solazio:
## Done

- Add dummy data to python and use it in templates
- Add 500 and 404 pages

## QA

- Homepage and `/charm` page should look the same but with some data come from the dummy data in python

## Issue / Card

Fixes #36


webapp/app.py
```diff
@@ -25,15 +25,189 @@ def utility_processor():
...
+            {
+                "height": "None",
+                "type": "screenshot",
+                "url": "https://dashboard.snapcraft.io/site_media/appmedia/"
+                "2018/04/2018-04-26-161203_1126x899_scrot_yBFL55L.png",
+                "width": "None",
+            },
+        ],
+        "videos": [],
+        "publisher_snaps": "None",
```
Lukewh: I *think* we should remove the following 3 lines.
`publisher_snaps`, `publisher_featured_snaps` and `has_publisher_page` as these are overrides we add on the frontend that won't be applicable to charmhub. This might change some of the checks you do in the templates!

Issue #47, opened 2020-04-01
Details page add a dummy bundle entity type


Lukewh:
Currently the only entity is a charm so we can't check the bundle pages.

PR #48, opened 2020-04-02
Add charm and bundle dummy data to python

### LINKED ISSUE #44: The main sidebar filters should be generated from the context
We should pass through the items in the sidebar in the context of the page as it will be generated from the API eventually.

solazio:
## Done

- Add dummy data to python for a bundle to be used on the details page
- Generate side navigation on home page using dummy data from python

## QA

- Go to http://0.0.0.0:8045 and see the side nav looks as previously
- Go to http://0.0.0.0:8045charm and http://0.0.0.0:8045/the-great-bundle?type=bundle and see that everything looks the same as before

## Issue / Card

Fixes #44 #47

### DIFF
--- a/webapp/charmhub.py
+++ b/webapp/charmhub.py
@@ -10,3 +10,333 @@ config = Config(
         "details_regex_uppercase": "[A-Za-z0-9-]*[A-Za-z][A-Za-z0-9-]*",
     }
 )
+
+
+# Dummy data for home page
+mock_search_results = [
+    {
+        "developer_validation": "unproven",
+        "entity_type": "bundle",
+        "icon_url": "https://dashboard.snapcraft.io/site_media/appmedia/"
+        "2016/09/logo_-_256px.png",
+        "media": [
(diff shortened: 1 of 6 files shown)


solazio:
> Good stuff. I think we should move this dummy data into the `webapp/charmhub.py` and import them. That way we can start adding some middleware for the filtering and searching in preparation. What do you think?

Sounds good. That way `app.py` will be easier to read.

webapp/charmhub.py
```diff
@@ -10,3 +10,333 @@ def __init__(self, d):
         "details_regex_uppercase": "[A-Za-z0-9-]*[A-Za-z][A-Za-z0-9-]*",
     }
 )
+
+
+# Dummy data for home page
+search_results = [
```
Lukewh, line 16: sorry I should have thought about this in my last comment - we should probably prefix these with something like "mock_" so when real data starts coming in we can differentiate.

PR #63, opened 2020-04-15
Make the whole card clickable

### LINKED ISSUE #43: Should the whole card be clickable?
On the homepage and related charms - should the whole card be the link? At the very least the title should be.

Perhaps the publisher could apply the publisher filter?

solazio:
## Done

- Make the whole card clickable

## QA

- See the whole card is clickable

## Issue / Card

Fixes #43 

## Screenshots


anthonydillon:
Could we look into using a button or button classes instead of a link wrapping a card as this will give use accessible focus states.

PR #87, opened 2020-04-23
Improve card heights to avoid scrollbars on smaller screens

Lukewh:
## Done

- Don't set an explicit height, on smaller screens that breaks things

## QA

- Visit http://0.0.0.0:8045/store
- Resize your browser
- Things should look ok

### DIFF
--- a/templates/store.html
+++ b/templates/store.html
@@ -39,7 +39,7 @@
           </select>
         </div>
       </div>
-      <div class="row u-equal-height">
+      <div class="row">
       {% if results %}
         {% for package in results %}
           {% include "partial/_entity-card.html" %}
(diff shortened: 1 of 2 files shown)


templates/store.html
```diff
@@ -39,7 +39,7 @@ <h4 class="p-muted-heading u-float-left">{{ results|length }} items</h4>
           </select>
         </div>
       </div>
-      <div class="row">
+      <div class="row u-equal-height">
```
solazio, line 42: this doesn't seem to do anything?

Lukewh: I'm just trying to get my characters-per-file count up! Will remove, thanks!

PR #93, opened 2020-04-27
Update details page

### LINKED ISSUE #79: Implement basic JS functionality on details page
- [ ] Tabs
- [ ] Accordion
- [ ] Show more
- [ ] Channel map
- [ ] Table of contents

Lukewh:
## Done

- Add JS for most tabs
- Add dummy data to channel map
- Added JS for channel map
- Updated some of the tabs with better fake data

## QA

- Visit http://0.0.0.0:8045
- Click around
- use the tabs
- change to different sections in the table of contents
- Refresh the page
- Use the browser forward and back buttons

## Issue / Card

Fixes #79


static/sass/styles.scss
```diff
@@ -102,6 +102,30 @@ $theme-default-nav: dark;
 
 @include charmhub-p-series-tag;
 
+.p-channel-map {
+  position: absolute;
+  z-index: 12;
+
+  &__mask {
```
solazio, line 109: ca you please use `p-channel-map__mask` instead

templates/partial/_tab-actions.html
```diff
@@ -3,9 +3,9 @@
     <hr class="p-separator" />
     <h2>Actions</h2>
   </noscript>
-  <aside class="p-accordion" role="tablist" aria-multiselect="true">
+  <aside class="p-accordion" role="tablist" aria-multiselect="true" data-js="actions">
```
solazio, line 6: do we want to add the expanded accordion to the url or it is to much?

templates/partial/_tab-docs.html
```diff
@@ -1,57 +1,61 @@
-<div class="u-fixed-width p-tabs__content" id="docs">
+{% set section_slug = "docs" %}
+{% set section_title = "Docs" %}
+{% include "partial/_partial-documentation.html" %}
+
+{#<div class="u-fixed-width p-tabs__content" id="docs">
```
solazio, line 5: I guess this part is supposed to be removed now that you have an include above?

templates/partial/_tabs-navigation.html
```diff
@@ -10,7 +10,7 @@
           <a href="#configuration" class="p-tabs__link" tabindex="-1" role="tab"
             aria-controls="configuration">Configuration</a>
         </li>
-        {% if entity_type == "charm" %}
+        {% if package["package-type"] == "charm" %}
```
solazio, line 13: does this not work as `package.package-type` to keep it consistent?

static/js/public/details/tableOfContents.js
```diff
@@ -0,0 +1,123 @@
+class TableOfContents {
+  constructor(el, historyState) {
+    this.el = el;
+    this.instanceName = el.dataset.js;
+    this.tabs = {};
+    this.historyState = historyState;
+
+    const tabLinks = this.el.querySelectorAll("[role='tab']");
+
+    tabLinks.forEach((tabLink, i) => {
```
solazio, line 10: perhaps we should check if `tabLinks` exists first?

static/js/public/details/tabs.js
```diff
@@ -0,0 +1,122 @@
+class Tabs {
+  constructor(el, historyState, children) {
+    this.el = el;
+    this.tabs = {};
+    this.historyState = historyState;
+    this.children = children;
+
+    const tabLinks = this.el.querySelectorAll("[role='tab']");
+
+    tabLinks.forEach((tabLink, i) => {
```
solazio, line 10: perhaps we should check if `tabLinks` exists first? Also, I am not sure if anything has changed since, but in the past I was advised not to use `forEach` directly on DOM nodes but instead do `[].slice.call(this.el.querySelectorAll("[role='tab']"))`.

templates/partial/_tab-actions.html
Lukewh: The issue with that is you can have multiple open at once, so not sure how we would codify that into the URL without it being hideous

templates/partial/_tabs-navigation.html
Lukewh: "-" is a reserved character, we would need to change all "-"'s to "_"'s in the backend, which might be worth doing at some point.

static/js/public/details/tabs.js
Lukewh: I've added charmhubs first polyfill to cover this

PR #422, opened 2020-09-24
Override mobile menu nav visibility toggle

### LINKED ISSUE #369: Filtering / Sorting doesn't work on small screens (after using main nav)
When you open/close main navigation dropdown on small screen it throws JS error about sorting.
After that you cant sort or filter.

charm-filter

barrymcgee:
## Done

This fixes a bug where the mobile nav toggling was interfering with the charm filtering/sorting on the homepage as they both rely on the url for state by disabling the targetting via the url and toggling the nav with JS instead.

Note, this is not the ideal fix for this systemic problem.. it's more of a sticky plaster.

The code for the filtering/sorting is quite fragile as it uses both the instance variable `this._filters` and query params in the URL for state at different points. To make this more robust to deal with hashes in the url requires a significant refactor or rewrite which I don't have the bandwidth for at this time - I'm also conscious this is quite a significant bug currently on production.

I've created a separate issue for the code smell here: https://github.com/canonical-web-and-design/charmhub.io/issues/421

## QA

- Change resolution to small screen
- Toggle nav
- Toggle filters
- Toggle nav
- Toggle filters
- Ensure no errors

## Issue / Card

Fixes #369

### DIFF
--- a/templates/partial/_navigation.html
+++ b/templates/partial/_navigation.html
@@ -6,8 +6,8 @@
           <img class="p-navigation__image" src="https://assets.ubuntu.com/v1/c3b8cc72-Charmhub_white_hex.svg" width="166" height="32" alt="Charmhub" />
         </a>
       </div>
-      <a href="#navigation" class="p-navigation__toggle--open" title="menu">Menu</a>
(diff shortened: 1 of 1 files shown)


solazio (approved):
Looks good :+1:. Thanks

PR #478, opened 2020-11-03
Add swap between channel easy

tbille:
## Done

- Refactor code for channel map
- Refactor code for details header page `get_package`
- Add ability to get the data for a specific channel
- Add back the channel map in header

## QA

- Use the staging API:
```
On file .env.local:

CHARMSTORE_API_URL=https://api.staging.snapcraft.io/
```
- `dotrun`
- http://0.0.0.0:8045/activemq?channel=stable
- Navigate in the pages and make sure that the channel doesn't change (see URL)
- http://0.0.0.0:8045/activemq
- Should work as usual too
- http://0.0.0.0:8045/activemq?channel=blablblablablabl
- Will default to the default channel (field from API) in most cases here `stable`

Unfortunately  there are no charms with different data between channels for now....

Fixes https://github.com/canonical-web-and-design/charmhub.io/issues/113

### DIFF
--- a/static/js/src/public/details/channelMap.js
+++ b/static/js/src/public/details/channelMap.js
@@ -23,27 +23,11 @@ const init = (packageName, channelMapButton) => {
   );
 
   const selectChannel = (channel, version) => {
-    const selected = document.querySelector(
-      `[data-channel-map-channel="${channelMapState.channel}"]`
-    );
-
-    selected.classList.remove("is-active");
-
-    channelMapState.channel = channel;
-    channelMapState.version = version;
-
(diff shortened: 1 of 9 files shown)


anthonydillon (changes requested):
Code looks good but one QA issue.

## QA
- If you are on a tab such as the action tab and change channel you jump back to the overview when I would expect to remain on the actions

tbille:
Good catch. Behaviour updated @anthonydillon

anthonydillon (approved):
Nice one :+1:

PR #498, opened 2020-11-04
Default empty readme to empty string

### LINKED ISSUE #497: Empty readme causes 500
It seems that this charm has no readme: hasan-test
We should default to empty in case this state happens.

tbille:
## Done

It seems that there is a way to have a charm without readme.
This is not a pattern that we want to encourage but currently the application 500.
Default the readme field to empty string to fix this.

## QA

     

- On file .env.local:
```
CHARMSTORE_API_URL=https://api.staging.snapcraft.io/
```
- `dotrun`
- http://0.0.0.0:8045/hasan-test

## Issue / Card

Fixes #497

### DIFF
--- a/webapp/store/views.py
+++ b/webapp/store/views.py
@@ -140,7 +140,9 @@ def details_overview(entity_name):
     channel_request = request.args.get("channel", default=None, type=str)
     package = get_package(entity_name, channel_request)
 
-    readme = package["channel_selected"]["revision"]["readme-md"]
+    readme = package["channel_selected"]["revision"].get(
+        "readme-md", "No readme available"
+    )
 
     # Remove Markdown comments
     readme = re.sub("(<!--.*-->)", "", readme, flags=re.DOTALL)


webapp/store/views.py
```diff
@@ -140,7 +140,7 @@ def details_overview(entity_name):
     channel_request = request.args.get("channel", default=None, type=str)
     package = get_package(entity_name, channel_request)
 
-    readme = package["channel_selected"]["revision"]["readme-md"]
+    readme = package["channel_selected"]["revision"].get("readme-md", "")
```
hatched, line 143: Lets make it clear what's going on here "No readme available" or "This charm has not provided a readme.md"

hatched (changes requested):

webapp/store/views.py
tbille: I agree, I was going to add some kind of notification. But for now I'll put `"No readme available"`

tbille: Done.

hatched (approved):

PR #534, opened 2020-11-08
Add structure data markup

tbille:
## Done

- Add structure data markup

## QA

- `dotrun`
- Check charm page header and make sure structure data seems correct
- Test with a tool like https://search.google.com/test/rich-results

## Issue / Card

Fixes #534


jkfran (approved):
This is super cool!

PR #538, opened 2020-11-09
Get all OS for a charm

tbille:
## Done

- On the homepage
- It will show all the os for a charm (for now there is no charm with multiple os...)

## QA

- `dotrun`
- http://0.0.0.0:8045
- should work as usual


jkfran (approved):
LGTM!

PR #547, opened 2020-11-10
Prefix charms install instructions

### LINKED ISSUE #546: Add prefix `cs:` to install instructions
Some charms will have the ability to be installed without `cs:`
The store API will provide a field to determine this.
As a tmp solution we can have the list hardcoded in the frontend

tbille:
## Done

Some charms will have the ability to be installed without `cs:`
All the rest of charms will have `cs:` in their install instructions

## QA

- `dotrun`
- http://0.0.0.0:8045/jenkins
- Make sure every charm has `cs:<charm>` in their install instruction
- Locally if you add a charm in the array `CS = ["jenkins"]` then the charm jenkins shouldn't be prefixed by `cs:` 

## Issue / Card

Fixes #546

### DIFF
--- a/templates/details/_details-header.html
+++ b/templates/details/_details-header.html
@@ -65,7 +65,7 @@
           </div>
         </div>
         {% include "partial/_channel-map.html" %}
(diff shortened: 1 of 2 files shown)


steverydz (approved):
LGTM :+1:

PR #602, opened 2020-11-12
Implement static features list and platform filtering

anthonydillon:
## Done
Implement static features list and platform filtering

**Needs the featured charm data**

## QA
- Check the set of static cards appear instantly and the count updates when the rest load
- Change the platform and see that the featured charms are replaced with a filtered list quickly
- See that the list count is correct
- Change the platform to All
- Check you see them all and the count is rendered correctly

## Issue / Card

Fixes https://github.com/canonical-web-and-design/charmhub.io/issues/584


steverydz (approved):
LGTM :+1:

PR #617, opened 2020-11-12
Added missing field for libraries API

jkfran:
## Done

Added missing field for libraries API

## QA

- There no libraries right now on staging or prod, but this field is on the response schema


solazio (approved):
Looks good :+1:. Thanks

PR #628, opened 2020-11-13
Proposing a clean up of the filter component

anthonydillon:
## Done
While looking into the checkbox jumping issue I noticed there are a lot of bespoke styling in the filters section of the site so I would like to propose this clean up of the filter component to use Vanilla much more so we don't get more difficult issues like this.

## QA
- Check that the filter section looks good
- Check it works as before


steverydz:
Looks a bit weird on mobile:

Screenshot from 2020-11-13 10-55-21

I was just looking at https://github.com/canonical-web-and-design/charmhub.io/issues/600 which touches on this too.

It's too complicated having two methods of handling the filters and platforms. Maybe at least we can keep the platforms visible as a select?

steverydz:
Something closer to this:

Screenshot from 2020-11-13 11-16-23

this way we don't have duplicated logic

steverydz (approved):
LGTM :+1:

PR #631, opened 2020-11-13
Disable actions until data is ready

steverydz:
## Done

Filtering of charms doesn't work until the data has loaded in the background. This PR addresses that by disabling all actions until the data is available.

## QA

- Check that the platform switcher, category filters and more operators button are all disabled until the data loads


jkfran (approved):
Looks good to me! 👍🏻

PR #641, opened 2020-11-16
Filters JS refactor

### LINKED ISSUE #554: Refactor filters JS to reduce complexity

steverydz:
## Done

Refactored JS for filters, including removing unused code and renaming functions and variables for improved readability.

## QA

- Play with the filters and platform switcher
- Check that it still works as it does on production

## Issue / Card

Fixes #554

### DIFF
--- a/templates/partial/_filters.html
+++ b/templates/partial/_filters.html
@@ -83,7 +83,7 @@
   </div>
 
   <div class="u-align--center" style="margin-top: 1rem;">
-    <button type="button" class="p-button--positive" id="more-operators" disabled>See all operators</button>
+    <button type="button" class="p-button--positive" id="show-all-charms" disabled>See all operators</button>
   </div>
 </div>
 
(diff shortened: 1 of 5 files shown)


jkfran (approved):

templates/partial/_filters.html
```diff
@@ -109,14 +109,6 @@ <h3 class="p-card__title p-heading--4 u-no-margin--bottom entity-card-title"></h
   </div>
 </template>
 
-<template id="no-results-message">
-  <div class="p-notification--information">
-    <p class="p-notification__response">
-      We couldn't find anything for your search<span class="search-query"></span>.
```
anthonydillon, line 115: Shouldn't we keep this message?

steverydz: It's never seen since the search results are now on their own page

jkfran: I thought about it as well but since we have now a different template for searches we don't need it.

anthonydillon: Ah of course :+1:

PR #643, opened 2020-11-16
Change login URL

jkfran:
## Done

We used to have two different logins:

1. As a publisher
2. To visit the entire website since it was private

Now we can rename `/publisher/login/` to `/login/`, this will make the site match with snapcraft.io.

## QA

**Login will not work on the demo service**
- Go to http://localhost:8045/login/
- Finish the login process, you will get redirected to `/charms`


steverydz (approved):
LGTM :+1:

PR #657, opened 2020-11-17
Improve the filter checkboxes

anthonydillon:
## Done
Improve the filter checkboxes

## QA
- See that when the filtered checkboxes are disabled the labels look it too


steverydz (approved):
LGTM :+1:

PR #660, opened 2020-11-17
Add download dropdown to library page

### LINKED ISSUE #508: Libraries page - download dialog
Add download dialog on download click - [design](https://app.zeplin.io/project/5f92a9530029eeac84ae9589/screen/5f92a9b491b20207df29ff12)

steverydz:
## Done

Add a download popup to the library page

## QA

- View the site locally in your web browser at: http://0.0.0.0:8045/fran-wordpress/libraries/test
- Hover over the download link
- Check that there is a popup with the command line instructions to download

## Issue / Card

Fixes #508

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -49,6 +49,7 @@ $theme-default-nav: dark;
 @include vf-p-table-mobile-card;
 @include vf-p-tabs;
 @include vf-p-tooltips;
+@include vf-p-contextual-menu;
 
 // Vanilla utilities
 @include vf-u-align;
@@ -160,6 +161,10 @@ $theme-default-nav: dark;
 @import "charmhub_metrics";
 
 @include charmhub-metrics;
+
+@import "charmhub_p-contextual-menu";
+
+@include p-charmhub-contextual-menu;
 
 @import "pattern_form-multiselect";
 
--- a/templates/details/libraries/libraries_layout.html
+++ b/templates/details/libraries/libraries_layout.html
@@ -81,7 +81,24 @@ than the latest.</span>
           <li class="p-inline-list__item">
             <ul class="p-inline-list has-vertical-separator">
               <li class="p-inline-list__item">
-                <a href="#" class="is-always-color-link"><i class="p-icon--download">Download</i> <span>Download</span></a>
+                <span class="p-contextual-menu--left">
+                  <button class="p-button--base p-contextual-menu__toggle" aria-controls="menu-3" aria-expanded="false" aria-haspopup="true">
+                    <i class="p-icon--download"></i> Download
+                  </button>
+                  <span class="p-contextual-menu__dropdown--wide" id="menu-3" aria-hidden="true">
+                    <span class="p-contextual-menu__group">
+                      <div class="p-contextual-menu__body">
+                        <p class="u-no-padding--top"><i class="p-icon--code"></i> Fetch library</p>
+                        <pre class="u-no-margin--bottom">charmcraft fetch-libs {{docstrings["name"]}}</pre>
+                      </div>
+                    </span>
+                    <span class="p-contextual-menu__group">
+                      <a class="p-contextual-menu__link--deep" href="/{{ library['charm-name'] }}/libraries/{{ docstrings['name'] }}/download" download>
+                        <i class="p-icon--download"></i> Download {{ docstrings['name'] }}.py
+                      </a>
+                    </span>
+                  </span>
+                </span>
               </li>
               <li class="p-inline-list__item">
                 <i class="p-icon--update">Last updated</i> 26 June 2019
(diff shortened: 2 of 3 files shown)


konf/site.yaml
```diff
@@ -11,9 +11,13 @@ env:
...
 
+  # # We have to use the staging APIs temporarily
+  # - name: CHARMSTORE_PUBLISHER_API_URL
+  #   value: https://api.staging.charmhub.io/
+
   # We have to use the staging APIs temporarily
-  - name: CHARMSTORE_PUBLISHER_API_URL
-    value: https://api.staging.charmhub.io/
+  - name: CHARMSTORE_API_URL
+    value: https://api.staging.snapcraft.io/
```
jkfran, line 20: @steverydz I think this is the issue, you don't need to remove `CHARMSTORE_PUBLISHER_API_URL`, you need to have both.

```sugestion
  - name: CHARMSTORE_API_URL
    value: https://api.staging.snapcraft.io/

  - name: CHARMSTORE_PUBLISHER_API_URL
    value: https://api.staging.charmhub.io/
```

static/sass/styles.scss
```diff
@@ -349,3 +349,24 @@ h3 {
...
+  display: none;
+  left: -1rem;
+  max-width: 420px;
+  padding: $sp-medium;
+  position: absolute;
+  top: 100%;
+  z-index: 10;
+}
+
+.p-modal-tooltip:hover .p-modal-tooltip__body {
```
tbille, line 370: I don't think we want it on `hover`. It feels on click is cleaner, we don't have random stuff poping up when i move my mouse around. @danielmutis ?

danielmutis: Yes I agree @tbille only on click is better.

steverydz:
@danielmutis the demo won't work anymore for this but this gif shows you it working on click:

download_button

steverydz:
@danielmutis

templates/details/libraries/libraries_layout.html
```diff
@@ -81,7 +81,23 @@ <h3>{{ docstrings["name"] }}</h3>
...
-                <a href="#" class="is-always-color-link"><i class="p-icon--download">Download</i> <span>Download</span></a>
+                <div class="p-modal-tooltip">
+                  <button class="p-button--base is-always-color-link has-icon u-no-margin--bottom" id="download-modal-button" aria-controls="download-modal">
+                    <i class="p-icon--download"></i>
+                    <span>Download</span>
+                  </button>
+                  <div class="p-modal-tooltip__wrapper u-hide" id="download-modal">
+                    <div class="p-modal-tooltip__body" aria-controlledby="download-modal-button">
+                      <p class="u-no-padding--top"><i class="p-icon--code"></i> Fetch library</p>
+                      <pre class="u-no-margin--bottom">charmcraft fetch-libs charms/{{library["charm-name"]}}/v{{library["api"]}}/db.py</pre>
```
tbille, line 92: This should stay: `charmcraft fetch-libs test`

The name of the library should be given by the API no need to build it.

templates/details/libraries/libraries_layout.html
```diff
@@ -81,7 +81,23 @@ <h3>{{ docstrings["name"] }}</h3>
           <li class="p-inline-list__item">
             <ul class="p-inline-list has-vertical-separator">
               <li class="p-inline-list__item">
-                <a href="#" class="is-always-color-link"><i class="p-icon--download">Download</i> <span>Download</span></a>
+                <div class="p-modal-tooltip">
```
tbille, line 84: Looks nice. Can we make it a bit bigger, so we can see all the command.

And click out of the modal should close it.

templates/details/libraries/libraries_layout.html
steverydz: done

templates/details/libraries/libraries_layout.html
steverydz: If it goes bigger it will go offscreen to the right. The alternative would be to extend it to the left but that will disassociate it from the button.
 
> And click out of the modal should close it.
I'd agree if this was actually a modal but it's a panel related to that button. @danielmutis do you have any thoughts on these points?

danielmutis:
@steverydz the grey background of the "Download library" section is just a hover state, if idle it  be white.
Can we also include the name of the library (i.e. db.py) rather than the word "library".

steverydz:
> @steverydz the grey background of the "Download library" section is just a hover state, if idle it be white.
> Can we also include the name of the library (i.e. db.py) rather than the word "library".

@danielmutis I've updated the text for the link, but why would we have such a large background area hoverstate for this link in particular. I can't see that we use this anywhere else?

Screenshot from 2020-11-24 14-21-08

steverydz:
@danielmutis 
Screenshot from 2020-11-24 14-40-24

steverydz:
@solazio ready for code review

anthonydillon (commented):
Out of curiosity, does a contextual menu not get your 85% of the way there?

steverydz:
> Out of curiosity, does a contextual menu not get your 85% of the way there?

@anthonydillon yes, you're right - I'll update it

steverydz:
@solazio this is now ready for code review (for real this time) :grin:

templates/details/libraries/libraries_layout.html
```diff
@@ -101,4 +118,68 @@ <h3>{{ docstrings["name"] }}</h3>
       </div>
     </div>
   </div>
+{% endblock %}
+
+{% block page_scripts %}
+  <script>
```
solazio, line 124: Not a big deal, but I would rather put this JS in the libraries JS file https://github.com/canonical-web-and-design/charmhub.io/tree/master/static/js/src/public/details/libraries

solazio (approved):
Just a suggestion in the code. Thanks

templates/details/libraries/libraries_layout.html
steverydz: I agree that we should have less JS in templates. I think this is a larger discussion though, as we have this task to deal with https://github.com/canonical-web-and-design/snapcraft.io/issues/1245 which involves us moving away from the pattern we currently have for loading scripts.

PR #690, opened 2020-11-25
Add summary and description to the charm overview page

solazio:
## Done

- Add summary and description to the charm overview page

## QA

- View the site locally in your web browser at: http://localhost:8045/discourse
- See there is a summary and a description at the top of the page

## Issue / Card

Fixes https://github.com/canonical-web-and-design/charmhub.io/issues/688

## Screenshots

image

### DIFF
--- a/static/js/src/libs/truncate-string.js
+++ b/static/js/src/libs/truncate-string.js
@@ -1,0 +1,28 @@
+/**
+ * Truncates a string to a word break.
+ * @param {string} str The string to be truncaed.
+ * @param {number} len The length that the string will be truncated to.
+ * @param {string} append This is optional.
+ * @returns {string} The truncated string.
+ */
+function truncateString(str, len, append = "...") {
+  let newLength;
+
+  if (str.indexOf(" ") + append.length > len) {
+    return str; //if the first word + the appended text is too long, the function returns the original String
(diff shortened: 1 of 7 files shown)


anthonydillon (changes requested):
I think we need to limit or address bad data before merging this. For example, https://charmhub-io-690.demos.haus/postgresql-k8s

solazio:
> I think we need to limit or address bad data before merging this. For example, https://charmhub-io-690.demos.haus/postgresql-k8s

I agree that looks bad, but is the only one that I could find and is like that. We can limit the summary length to 90 (or whatever number makes sense) characters. What do you think @danielmutis ? Or can we update the charm metadata ourselves?

danielmutis:
The summary should be limited to 100 characters as we calculated on Monday. For some reason in the cards for postgresql we show the description and not the summary?

cristinadresch:
@danielmutis do we need a new design for this?

steverydz (approved):
LGTM :+1:

danielmutis:
I think this is a good way of managing bad data, but we shouldn't merge this until they have had a chance to fix the new charms that have bad info here.

anthonydillon (approved):

PR #700, opened 2020-11-30
WIP staging demo

tbille:
# DON'T MERGE THIS IS FOR DEMO PURPOSE

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -8,6 +8,12 @@ env:
       key: maintenance
       name: charmhub
 
+  - name: CHARMSTORE_API_URL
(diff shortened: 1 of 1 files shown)


steverydz:
@tbille @jkfran do we still need this open?

tbille:
CC @Roadmaster are you still using this demo to test the APIs?

Roadmaster:
We use it from time to time, we don't have a lot of API changes coming so we might not need it in the immediate future but we do occasionally need it.

If you need to kill it that should be OK, assuming that you can stand up a new one if needed :)

PR #770, opened 2021-01-08
749 bundle overview page

### LINKED ISSUE #749: Charm bundles: Overview page

solazio:
## Done
- _Update bundle icon on home page_
- _Update bundle icon on search page_
-  _Add svg topology to bundle overview page_
- _Show only the `Overview` tab for bundles_

## How to QA
- Visit the demo in your browser, link commented below
  - Or you can run the project locally using the [dotrun](https://snapcraft.io/dotrun) snap with `$ dotrun` and view it in your web browser at: http://localhost:8045/
  - If your demo doesn't work, read more about the [demoservice](https://discourse.ubuntu.com/t/demoservice/16862)
- Note that bundles are displayed and the icon looks right
- Click on one bundle and see the topology on the overview page
- Search for `kubeflow` and see the search results look right
- Go to `/prometheus` and open the console, and see there are no errors

## Issue / Card
Fixes 

## Screenshots


webapp/store/views.py
```diff
@@ -186,6 +187,8 @@ def details_overview(entity_name):
 
     readme = md_parser(readme)
     readme = decrease_headers(readme)
+    if package["type"] == "bundle":
+        readme = inject_topology_image(readme, package["name"])
```
tbille, line 191: I think this is not necessary for now.
Let's keep the page simple. We can't get the topology now, let's just show the readme.

tbille:
I am a bit confused by this PR. I feels we don't really need a feature branch for bundles. We work in iteratively, we should be able to land feature by feature.

This PR adds quite a lot of logic:
- renaming a variable across backend and templates
- Add support for topology
- adding support for overview page
- and more

It feels this PR should be split a bit more.

We can continue on this one now that this is done. So from what I see in here, this should be udpated:
- [ ] Remove adding fake topology
- [ ] Remove tabs (actions, docs, configure) and only keep overview for now
- [ ] This icon generated seems to small compare to the text
image

clagom:
On the Dashboard there's a small version of the topology that can be opened up for a bigger view. I wonder if we want to explore something in this sense for a next iteration on the page @danielmutis

danielmutis:
@clagom so this is the same image but more compact? I think the main issue with the bundle page is how big the topology is, because it pushes the content down. What do you think about maybe hiding part of the image?

clagom:
> @clagom so this is the same image but more compact? I think the main issue with the bundle page is how big the topology is, because it pushes the content down. What do you think about maybe hiding part of the image?

Yes I saw the issue of having this big topology on top pushing down the content, too prominent. So I wondered if we can reuse the smaller topology window that we have in the Dashboard (different topology, just an example):
Screenshot 2021-01-13 at 10 03 33
Screenshot 2021-01-13 at 10 03 39

Then users can open it up if they want to have a better look (especially for complex/big bundles).

PR #817, opened 2021-02-15
Fix JS filters when missing categories

### LINKED ISSUE #816: Issues with filters and API changes
image

jkfran:
## Done

For some reason, some categories are now missing from the API, we'll talk with the store team about this (@tbille )

Our JS code was failing until we rolled back to a previous version of the site. With these changes, the site will continue to work as usual, the category filters will be disabled for the missing categories.

## How to QA
- Check the filters are working as expected.
- Some categories are not available now.

## Issue / Card
Fixes #816

## Screenshots
image


tbille (approved):
QA+1 for me
We should add tests to this section in the future.

solazio (approved):
Looks good :+1:. Thanks

PR #831, opened 2021-02-22
WIP: Test typescript migration

solazio:
## Done
- _Migrate the project to using both javascript and typescript._

## How to QA
- Run the project locally using the [dotrun](https://snapcraft.io/dotrun) snap with `$ dotrun` and view it in your web browser at: http://localhost:8045/
- See it works with both `*.js` and `*.ts` files

## Issue / Card
Fixes #

## Screenshots


solazio:
@anthonydillon @nottrobin @hatched @barrymcgee @bartaz @tbille @pmahnke 
This a test PR for migrating charmhub.io to typescript. It enables the use of both JavaScript and TypeScript at the same time. The build time is a little bit longer because we have to run `tsc`.
In my opinion this is a pretty smooth move to TypeScript, and can be done file by file. In terms of "sctrictness" I think we should be as strict as possible with TypeScript. 
Please let me know your thoughts and if we should start migrating more projects to using TypeScript.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "ES2020", // default value is ES3
+    "strict": true,
+    // "strictNullChecks": false,
```
barrymcgee, line 6: Is there any reason to keep this config option if commented out?

barrymcgee:
Is there a parsing error here?

Screenshot 2021-02-22 at 09 40 17

tsconfig.json
solazio: No, it can be removed. I'll remove it.

barrymcgee:
Unrelated but this codebase is misconfigured for Prettier - have raised a separate issue here: https://github.com/canonical-web-and-design/charmhub.io/issues/832

templates/publisher/publicise/github_badges.html
```diff
@@ -1,5 +1,37 @@
 {% extends 'publisher/publicise/publicise_layout.html' %}
 
 {% block publicise_content %}
-  <h4>Github badges page</h4>
+  <h4>Promote your charm using an embeddable GitHub badge</h4>
```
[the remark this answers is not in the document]
solazio: No..I started from a branch that was supposed to be merged by now, but it's not. I'll remove these changes to avoid confusion.

tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "module": "commonjs",
+    "target": "ES2020", // default value is ES3
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true,
+    "allowSyntheticDefaultImports": true,
+    "allowJs": true,
+    "isolatedModules": true,
+    "noImplicitAny": false,
```
hatched, line 12: I think this is the default.

hatched (approved):
Looks good 👍

package.json
```diff
@@ -3,12 +3,13 @@
...
     "clean": "rm -rf node_modules yarn-error.log css static/css *.log *.sqlite _site/ build/ .jekyll-metadata .bundle",
-    "watch": "watch -p 'static/sass/**/*' -p 'static/js/**/*.js' -c 'yarn run build'",
+    "watch": "watch -p 'static/sass/**/*' -p 'static/js/**/*.js' -p 'static/js/**/*.ts' -c 'yarn run build'",
     "watch-js": "webpack --watch",
     "watch-scss": "watch -p 'static/sass/**/*.scss' -c 'yarn run build-css'",
     "build": "yarn run build-css && yarn run build-js",
     "build-css": "sass static/sass/styles.scss static/css/styles.css --load-path=node_modules --style=compressed && postcss --map false --use autoprefixer --replace 'static/css/**/*.css'",
-    "build-js": "yarn run build-js-bundle",
+    "type-check-js": "tsc",
+    "build-js": "yarn run build-js-bundle && yarn run type-check-js",
```
nottrobin, line 12: Shouldn't you type check first?

package.json
```diff
@@ -3,12 +3,13 @@
   "license": "LGPL-3.0-or-later",
   "scripts": {
     "clean": "rm -rf node_modules yarn-error.log css static/css *.log *.sqlite _site/ build/ .jekyll-metadata .bundle",
-    "watch": "watch -p 'static/sass/**/*' -p 'static/js/**/*.js' -c 'yarn run build'",
+    "watch": "watch -p 'static/sass/**/*' -p 'static/js/**/*.js' -p 'static/js/**/*.ts' -c 'yarn run build'",
```
nottrobin, line 6: What's the difference between a `.js` and a `.ts` file? Can you simply take a plain `.js` file, whack `.ts` on the end and compile it with typescript, without any other changes?

Because if so, I'd suggest we just rename them all now and compile them all with typescript, and rip out `.js` support altogether.

hatched: There are some minor changes that may be required but I'd prefer we don't do what you're proposing as that's the bare minimum. I'd prefer that when something gets converted to TS we have made the conscious effort to go through the file and truly "convert" it.

package.json
solazio: I think that's a good idea. Will do the type check before I bundle.

tsconfig.json
solazio: I'll double check and remove them if that is the case.

package.json
solazio: If we go with the approach of strict TypeScript rules, then you cannot just change `*.js` to `*.ts` because it will fail. If you change the file to `*.ts` you will have to deal with all the errors and amend the file so it passes the type check.

I think we could do that, and then incrementally make the rules stricter, however I don't like that approach as you will have to fix all the files when you change the strictness. Also it will create inconsistencies between projects, as some will have stricter rules than others overtime.

tsconfig.json
solazio: you were right, that is the default for `allowSyntheticDefaultImports`. I'm removing it.

package.json
nottrobin: Ok thanks for explaining.

tbille:
oooh typescript. It seems it was pretty light to add typescript support to the site. That's nice.

I guess with this PR we want to show how simple it is to add typescript support to all of our sites. I have a few questions about this:

- are we worried to have issues when integrating with other libraries that are **not** using typescript? Are there cases where it might not work?
- What is the release model for typescript, should we worry that this will add an extra dependency that might be more complex to upgrade across our projects? We see already how it is to upgrade webpack to major versions across all the sites, what will it be for typescript?
  - Should we pin typescript in the same way that we "pin" python?
- Can we see a performance gain with the transpiled JS? Since we are not using types now, are we seeing some lower performances because of type checking on the transpiled JS, or is the typing mostly useful for the linters and the IDE autocompletion?

I can't make my mind on the use of typescript for mostly JS that does DOM manipulations. Is that over engineering and just more work for the teams that maintain the dependencies across sites? Or thanks to typing we might be able to write more "predictable" software leading to less bug maintenance?

solazio:
- I am not aware of any problems with any libraries if you use TypeScript. At the end of the day it is still JavaScript... on steroids.
- As any other dependency we can pin it. I don't think TypeScript will pose any difficulties in terms of upgrades, and it is not comparable with webpack.
- There is no performance improvement, on the contrary it will be slightly slower to build in our current setup, but I don't see that as a problem.
- Using TypeScript (even with small DOM manipulation projects) will improve the quality of the code, catch some edge cases, and reduce the number of bugs in our code.

hatched:
> are we worried to have issues when integrating with other libraries that are **not** using typescript? Are there cases where it might not work?

TS is always compiled to JS before execution so it will always be able to integrate with other JS. The catch is that sometimes it can be a challenge if you're trying to integrate with a library that doesn't have types defined as you may have to spend a bit of time typing that yourself.
 
> What is the release model for typescript, should we worry that this will add an extra dependency that might be more complex to upgrade across our projects? We see already how it is to upgrade webpack to major versions across all the sites, what will it be for typescript?

TS maintains fairly strict backwards compatibility standards - we haven't had any issues in other projects upgrading yet.

> Should we pin typescript in the same way that we "pin" python?

It should be handled as any other JS dependency upgrade.

> Can we see a performance gain with the transpiled JS? Since we are not using types now, are we seeing some lower performances because of type checking on the transpiled JS, or is the typing mostly useful for the linters and the IDE autocompletion?

The big gain is in development speed (once familiar) and in program stability, as it's not a runtime language there is no performance gain (or at least none that's of any consequence).

> I can't make my mind on the use of typescript for mostly JS that does DOM manipulations. Is that over engineering and just more work for the teams that maintain the dependencies across sites? Or thanks to typing we might be able to write more "predictable" software leading to less bug maintenance?

It's true that using TS when doing basic DOM manipulations is a bit overkill but it's nice when actually writing the code as the compiler/editor will warn you when you're using the api's incorrectly. It also provides very good auto-completion.

Projects also have a way of becoming more complex over time and having TS already setup will help as it slowly grows in complexity. There will be a bit of a learning curve but once you're familiar with it there is a marketable improvement.

One of the good things about TS is that JS is valid TS (although our rules invalidate that statement a bit when it comes to catching some bad habits) so if we have someone that's coming on board that's just getting up to speed with JS they can still be productive and the typed enhancements can be added in follow-ups...of course this isn't ideal but it is possible :)

tbille:
@solazio 

> - As any other dependency we can pin it. I don't think TypeScript will pose any difficulties in terms of upgrades, and it is not comparable to webpack.

I agree that it's not comparable to webpack. In my opinion it's more complex.

Where webpack is mostly a configuration based system, Typescript is a full layer on top of the JS to "improve" it. This means that if we drop it we need to rewrite quite a lot of code instead of just updating some commands or configuration (I am being not optimistic just for the example).

> - There is no performance improvement, on the contrary it will be slightly slower to build in our current setup, but I don't see that as a problem.

I guess this will mostly visible on the day to day work.

> Using TypeScript (even with small DOM manipulation projects) will improve the quality of the code, catch some edge cases, and reduce the number of bugs in our code.

:+1: 

@hatched 

> TS maintains fairly strict backwards compatibility standards - we haven't had any issues in other projects upgrading yet.

:+1: 

> The big gain is in development speed (once familiar) and in program stability, as it's not a runtime language there is no performance gain (or at least none that's of any consequence).

:+1: 

> Projects also have a way of becoming more complex over time and having TS already setup will help as it slowly grows in complexity. There will be a bit of a learning curve but once you're familiar with it there is a marketable improvement.

I think on charmhub it would make sense to integrate TS especially when we will be working on the publisher section of the site. 

What you say makes complete sense, being able to code in the best conditions is great but I still have doubts into adding TS into our "base stack". I'll take some time to read a bit more about TS.

That said I am happy to go ahead with charmhub and I hope in the future snapcraft.

hatched:
> Where webpack is mostly a configuration based system, Typescript is a full layer on top of the JS to "improve" it. This means that if we drop it we need to rewrite quite a lot of code instead of just updating some commands or configuration (I am being not optimistic just for the example).

To assuage those concerns a bit...In the event that we were to drop TS, the TS compiler actually ouputs really clean legible JS that we could commit in leu of the TS code and carry on.

tbille:
Closing this one. We will work on this on a planned a bigger scale

PR #849, opened 2021-03-02
1821 add args to library functions

solazio:
## Done
- _Display function arguments on library docstring page._

## How to QA
- Run the project locally using the [dotrun](https://snapcraft.io/dotrun) snap with `$ dotrun` and view it in your web browser at: http://localhost:8045/fran-wordpress/libraries/test
- See that functions have arguments displayed inside the brackets if they exist

_Will only work once https://github.com/canonical-web-and-design/canonicalwebteam.docstring-extractor/pull/8 is merged_

## Issue / Card
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1821

## Screenshots
image


webapp/store/views.py
```diff
@@ -318,6 +318,15 @@ def details_library(entity_name, library_name):
...
 
+    function_args = {}
+    for text in re.split("\n+", library["content"]):
+        stripped_text = text.strip()
+        if stripped_text.startswith("def"):
+            stripped__arr = stripped_text.split("(")
+            func_name = stripped__arr[0].split(" ")[1]
+            func_args = stripped__arr[1].split(")")[0]
+            function_args[func_name] = func_args
+
```
tbille, line 329: Is the parser not handling that for you?

solazio: unfortunately not

tbille: I think this logic should be added in here https://github.com/canonical-web-and-design/canonicalwebteam.docstring-extractor/blob/main/canonicalwebteam/docstring_extractor/__init__.py#L24

This will make this code available in case we start using this parser in other projects.

solazio: @jkfran but do I have access in the `docstring_extractor` to the file content?

jkfran: It needs to be done in the `docstring_extractor`. Yes, you have access to the code, instead of doing it like this, using `ast` will be the neater way.

jkfran (approved):
Well done Ovi, LGTM 👍🏽

PR #858, opened 2021-03-10
Fix hidden content under the sticky nav bar on actions page

### LINKED ISSUE #857: Clicking an action name jumps page surprisingly
https://charmhub.io/ceph/actions

Click one of the action names in the left column jumps you to the action description below the header so you don't have any context as to where you are.

Screen Shot 2021-03-09 at 2 50 22 PM

I expected to see the heading to provide some context, highlighting the selected element and the action list should probably also follow you so you don't have to scroll all the way back up again...or maybe just add a "back to top" link for the selected element.

Something like this quick mockup...

Screen Shot 2021-03-09 at 2 54 37 PM

solazio:
## Done
- _Fix hidden content under the sticky nav bar on actions page._

## How to QA
- Go to https://charmhub-io-858.demos.haus/ceph/action and click some items in the side navigation
- See it takes you to the right section and you can see the title of that section

## Issue / Card
Fixes #857 

## Screenshots

### DIFF
--- a/templates/details/actions.html
+++ b/templates/details/actions.html
@@ -79,4 +79,8 @@
 
 {% block details_scripts %}
   <script src="{{ versioned_static('js/dist/highlight-nav.js') }}" defer></script>
+  <script>
+      // Fix to see the content under the sticky nav when you navigate using #id
+      window.addEventListener("hashchange", () => { scrollBy(0, -65) })
+  </script>
 {% endblock %}


templates/details/actions.html
```diff
@@ -79,4 +79,8 @@ <h4>No actions have been added yet</h4>
 
 {% block details_scripts %}
   <script src="{{ versioned_static('js/dist/highlight-nav.js') }}" defer></script>
+  <script>
+      // Fix to see the content under the sticky nav when you navigate using #id
+      window.addEventListener("hashchange", () => { scrollBy(0, -45) })
```
tbille, line 84: would be nice to add a bit more space so the title is not attached directly to the navigation
image

Something like this:
image

tbille (approved):

PR #890, opened 2021-03-21
Fix missing use of variable

### LINKED ISSUE #889: Argument wrong in the function parentesis on librairy page
This seems to only happen to functions.

See the `function(arg)`. The `arg` should actually be `field`

tbille:
## Done
- The variable in the loop wasn't use for the parameters of a function in the libraries tab
- Remove extra space before the first argument name  

## How to QA
- On staging api, add this to the `.env.local`:
```
# Staging API for publisher:
CHARMSTORE_PUBLISHER_API_URL=https://api.staging.charmhub.io/
CANDID_API_URL=https://api.staging.jujucharms.com/identity/

# Staging API for consumer:
CHARMSTORE_API_URL=https://api.staging.charmhub.io/
```
- http://0.0.0.0:8045/toto-wordpress/libraries/libwithdocs
- Check that the functions have the correct name of arguments (not just `arg`)

## Issue / Card
Fixes #889

### DIFF
--- a/templates/details/libraries/docstring.html
+++ b/templates/details/libraries/docstring.html
@@ -47,10 +47,7 @@
           <h4 id="{{ group['name'].lower() }}" class="p-docstring-block__heading">
             {% if group["type"].lower() == "function" %}
               <span class="token keyword">def</span>
-              <span class="token function">{{ group["name"] }}(</span>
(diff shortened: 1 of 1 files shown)


solazio (approved):
Looks good :+1:. Thanks

PR #916, opened 2021-04-06
Add nightly linkchecker

### LINKED ISSUE #672: Add nightly linkchecker
We should add a nightly linkchecker to avoid broken links.

The amount of charms is still manageable to run linkchecker on the entire site.

solazio:
## Done
- _Add nightly linkchecker._

## Issue / Card
Fixes #672

### DIFF
--- a/.github/workflows/linkchecker.yaml
+++ b/.github/workflows/linkchecker.yaml
@@ -1,0 +1,24 @@
+name: Run linkchecker on site
+
+on:
+  schedule:
+    - cron:  '0 14 * * 1'
+
+jobs:
+  run-linkchecker:
+    if: github.repository == 'canonical-web-and-design/charmhub.io'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout master
+      - uses: actions/checkout@v2
+
+      - name: Install linkchecker
+        run: sudo pip install LinkChecker
+
+      - name: Run linkchecker
+        run: linkchecker --ignore-url "http://localhost*" --ignore-url "https://assets.ubuntu.com$" --check-extern --no-warnings https://charmhub.io
+
+      - name: Send message on failure
+        if: failure()
+        run: curl -X POST -F "workflow=${GITHUB_WORKFLOW}" -F "repo_name=${GITHUB_REPOSITORY}" -F "action_id=${GITHUB_RUN_ID}" ${{ secrets.BOT_URL }}?room=%23web-team


.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+      - name: Install linkchecker
+        run: sudo pip install LinkChecker
+
+      - name: Run linkchecker
+        run: linkchecker --threads 2 --ignore-url "http://localhost*" --ignore-url "https://assets.ubuntu.com$" --check-extern --no-warnings https://charmhub.io
+
+      - name: Send message on failure
+        if: failure()
+        run: curl -X POST -F "workflow=${GITHUB_WORKFLOW}" -F "repo_name=${GITHUB_REPOSITORY}" -F "action_id=${GITHUB_RUN_ID}" ${{ secrets.BOT_URL }}?room=%23web-team
```
tbille, line 24: not sure about the ping on the webteam channel

solazio: @anthonydillon is using [docs](https://github.com/canonical-web-and-design/multipass.run/blob/4818bab04a16028ca67cca5367f6817a30fe73ee/.github/workflows/docs-links.yaml#L24) for Docs link check. Shall we create a separate channel for linkchecker notifications?

anthonydillon: Webteam would be fine as its information we should want to know. Creating a new room is likely for these notifications to be missed,

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+    steps:
+      - name: Checkout master
+      - uses: actions/checkout@v2
+
+      - name: Install linkchecker
+        run: sudo pip install LinkChecker
+
+      - name: Run linkchecker
+        run: linkchecker --threads 2 --ignore-url "http://localhost*" --ignore-url "https://assets.ubuntu.com$" --check-extern --no-warnings https://charmhub.io
```
anthonydillon, line 20: You could remove the `--threads 2` so it defaults to 10 to speed up the checks

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,24 @@
+name: Run linkchecker on site
+
+on:
+  schedule:
+    - cron:  '0 0 1,15 * *'
```
anthonydillon, line 5: 00:00 UTC is a strange time to run. Probably more efficient to run in main working hours.

anthonydillon (commented):
This is not set to nightly.

.github/workflows/linkchecker.yaml
solazio: fair enough, I'll run it at 12:00

.github/workflows/linkchecker.yaml
solazio: In that case I keep it as it is.

tbille: with the `--check-extern` if any of the charms publisher posts a 404 links we will always get this message until the publisher fixes it

solazio:
> This is not set to nightly.

@anthonydillon is set to run on the 1st and 15th of every month... because I definitely had in mind that it has to run fortnightly. I will change it to run every day.

anthonydillon:
> @anthonydillon is set to run on the 1st and 15th of every month... because I definitely had in mind that it has to run fortnightly. I will change it to run every day.

I don't mind which you use but the code is fortnightly and the commit says nightly.

tbille (approved):
as discussed in standup let's see how it goes.

PR #921, opened 2021-04-12
Use title of an operator instead of id when defined

tbille:
## Done
- use title of a charm if it is defined

## How to QA
-  `dotrun`
- There no charms currently with title set. Check for now if the homepage and the details page are still showing the charm names
- http://0.0.0.0:8045/charmcraft-prometheus
- http://0.0.0.0:8045/


solazio (approved):
Looks good :+1:. Thanks

PR #928, opened 2021-04-14
Enable bundles

### LINKED ISSUE #927: Enable bundles on charmhub.io
To enable bundles on charmhub.io we should:
- [x] remove the topology that comes from the previous store
- [x] Fix the platform on the homepage
- [x] Fix the bundle pages that don't currently work (eg. http://0.0.0.0:8045/canonical-kubernetes-canal )

solazio:
## Done
- _Enable bundles on homepage._

## How to QA
- Visit https://charmhub-io-928.demos.haus and click on "see all operators" button
- You should see some bundles as well as charms (like kubeflow) 
- Click around and see everything works fine, including the search

**Does not work on the staging API!!!**

## Issue / Card
Fixes #927

### DIFF
--- a/static/js/src/public/store/buildPackageCard.js
+++ b/static/js/src/public/store/buildPackageCard.js
@@ -1,5 +1,25 @@
 import { truncateString } from "../../libs/truncate-string";
 
+function buildPlatformIcons(entityCardIcons, altText, srcText, text) {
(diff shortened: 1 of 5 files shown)


tbille (approved):
LGTM.
Would be curious to know what is happening with staging.

PR #930, opened 2021-04-15
Stop using platform, use base instead

tbille:
## Done
- Platforms are going to be dropped
- they are getting replaced with base
- This PR replaces every occurance of platform with base.

## How to QA
- `dotrun`
- http://0.0.0.0:8045/
- http://0.0.0.0:8045/containers-metallb-speaker
- Make sure all the filters are working

### DIFF
--- a/static/js/src/public/store/buildPackageCard.js
+++ b/static/js/src/public/store/buildPackageCard.js
@@ -90,50 +90,47 @@ function buildPackageCard(entity) {
   }
 
   const entityCardIcons = clone.querySelector(".package-card-icons");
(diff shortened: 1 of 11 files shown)


solazio:
I think we should have the two things below consistent: both "platform" or both "base"
image

tbille:
@solazio updated.

solazio:
sorry @tbille, I also noticed this:
image

tbille:
@solazio updated :smile:

solazio (approved):
Looks good :+1:. Thanks

PR #936, opened 2021-04-21
Stop using format_slug in templates

### LINKED ISSUE #935: Display name casing not respected
I've just uploaded and released https://charmhub.io/postgresql-k8s using charmcraft, and in metadata.yaml I see the following:
```
name: postgresql-k8s
display-name: PostgreSQL
```
However, on the charmhub page, the title is displayed as "Postgresql" rather than "PostgreSQL".

tbille:
## Done
- Stop using format_slug in templates
- If there is no result.title on a charm then use the format_slug function

## How to QA
- `dotrun`
- http://0.0.0.0:8045/
- https://0.0.0.0:8045/ceilometer (should be a capital C because it has no title)
- http://0.0.0.0:8045/postgresql-k8s (should be PostgreSQL because it has a title)

## Issue / Card
Fixes #935

### DIFF
--- a/webapp/store/logic.py
+++ b/webapp/store/logic.py
@@ -238,7 +238,7 @@ def add_store_front_data(package, details=False):
     if "title" in package["result"] and package["result"]["title"]:
         extra["display-name"] = package["result"]["title"]
     else:
-        extra["display-name"] = package["name"]
+        extra["display-name"] = format_slug(package["name"])
 
     if details:
         extra["metadata"] = yaml.load(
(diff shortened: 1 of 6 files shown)


webapp/store/logic.py
```diff
@@ -365,3 +365,23 @@ def filter_charm(charm, categories=["all"], base="all"):
...
+def format_slug(slug):
+    """Format slug name into a standard title format
+    :param slug: The hypen spaced, lowercase slug to be formatted
+    :return: The formatted string
+    """
+    # TODO Remove once we have category on store side
+    if slug.lower() == "ai/ml":
+        return slug.upper()
+    if slug.lower() == "openstack":
+        return "OpenStack"
```
jkfran, line 379: I think this is not needed anymore

tbille: good catch. I removed this section

jkfran (approved):
LGTM

PR #941, opened 2021-04-21
Implement the new navigation

solazio:
## Done
- _Implement the new navigation._

## How to QA
- Visit https://charmhub-io-941.demos.haus
- See the navigation looks and behaves the same as the one on http://juju.is/

## Issue / Card
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/1997


tbille (approved):
QA+1

solazio:
I still need to add the search bar

static/js/src/base/navigation.js
```diff
@@ -87,4 +100,38 @@ document.addEventListener("click", function (event) {
...
+      );
+      // ask the observer to observe the position of the observerEl - i.e. if it's on/off screen
+      observer.observe(observerEl);
+    } else {
+      throw new Error(
+        `${selector ? selector : "#navigation"} is not a valid element!`
+      );
+    }
+  });
+};
```
sowasred2012, line 133: I think this is being transpiled with babel - is that right? ES5 and ES6 syntax is being mixed in this one function: the function itself is a `const` and uses an arrow function expression, but the body of the function contains the use of `var` and traditional function expressions, along with ES6 style string interpolation.

If it's being transpiled, it doesn't matter whether it's ES5 or ES6, but I think the file should be consistent one way or the other.

sowasred2012 (commented):
Looks good - will hold off on a code +1 until the search bar is added, but had one comment.

static/js/src/base/navigation.js
solazio: Yes, it is!you are right. this code is transpiled, but on juju.is is not. I'll have to amend the file there. I didn't see the elephant in the room 💯

sowasred2012 (approved):
LGTM

PR #958, opened 2021-04-27
Fix index issue when no icons

tbille:
## Done
- When no icon is set there is an index issue
- When a charm has no icon it should return the default icon

## How to QA
- `dotrun` 
- http://0.0.0.0:8045/containers-metallb-speaker/icon
- should return default icon
- http://0.0.0.0:8045/prometheus/icon
- should return prometheus icon


hatched (approved):
🙏 much appreciated!

PR #960, opened 2021-04-28
Use deployable-on field

### LINKED ISSUE #954: Use result.deployable-on to find out K8s charms
find results from the charmhub API (and info too) now support a `result.deployable-on` field. It will always return a list. If it contains "kubernetes" in the list, the charm is for k8s. Right now the only other option is an empty list for a non-k8s charm. But eventually it'll show, for example, whether a charm is a machine charm, or once we have universal charms, it might contain both machine and kubernetes. And maybe other things later on. TL;DR if you request it and it contains "kubernetes" it's a k8s charm. you can easily try it with the featured category

jkfran:
## Done
- Use new API field `result.deployable-on`

## How to QA
- Check https://charmhub-io-960.demos.haus/mattermost-charmers-mattermost the Kubernetes icon should appear
- Check https://charmhub-io-960.demos.haus/postgresql the Ubuntu icon should appear
- Check filters are working in the main view.

## Issue / Card
Fixes #954

### DIFF
--- a/static/js/src/public/store/buildPackageCard.js
+++ b/static/js/src/public/store/buildPackageCard.js
@@ -96,7 +96,7 @@ function buildPackageCard(entity) {
   }
 
   const entityCardIcons = clone.querySelector(".package-card-icons");
-  if (entity.store_front.base === "kubernetes") {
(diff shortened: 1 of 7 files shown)


tbille (approved):
Looks good!

PR #972, opened 2021-04-29
Add framework labels to homepage cards

steverydz:
## Done
Added framework type labels to homepage cards

## QA
- Go to https://charmhub-io-972.demos.haus/
- Check that the homepage cards have labels for either "Operator framework" or "Reactive"
- Check that reactive has a tooltip

### DIFF
--- a/static/js/src/public/store/buildPackageCard.js
+++ b/static/js/src/public/store/buildPackageCard.js
@@ -89,6 +89,41 @@ function buildPackageCard(entity) {
     );
   }
 
+  const charmFrameworkTypeIcon = clone.querySelector(
+    "[data-js-charm-framework-type-icon]"
+  );
+
+  const charmFrameworkTypeText = clone.querySelector(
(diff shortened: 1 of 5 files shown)


spencerbygraves:
Thanks @steverydz, they look a little bit misaligned horizontally, and there should be 8px between icon and text. [Design for reference](https://zpl.io/aNxDqzQ).

solazio:
I don't know if you want to do this here, but we should update the placeholders as well. It requires an aditional rectangle in the svg to be aligned with the "reactive" text https://github.com/canonical-web-and-design/charmhub.io/blob/master/templates/store.html#L82
image

solazio (approved):
Looks good. I left a separate comment, but that can be addressed separately.

steverydz:
@solazio thanks. I've made a separate issue for the grey bar, so we can get this over the line: https://github.com/canonical-web-and-design/charmhub.io/issues/974

PR #990, opened 2021-05-04
Add hero typer to homepage

anthonydillon:
## Done
Add hero typer to homepage

## How to QA
- Visit the demo in your browser, link commented below
- See that the words in the title auto type out
- Check small screens reverts to a single word

### DIFF
--- a/static/js/src/libs/typer.js
+++ b/static/js/src/libs/typer.js
@@ -1,0 +1,128 @@
+var Typer = function (element) {
+  this.element = element;
+  var delim = element.dataset.delim || ",";
+  var words = element.dataset.words || "override these,sample typing";
+  this.words = words.split(delim).filter((v) => v); // non empty words
+  this.delay = element.dataset.delay || 200;
+  this.loop = element.dataset.loop || "true";
+  if (this.loop === "false") {
+    this.loop = 1;
+  }
+  this.deleteDelay =
+    element.dataset.deletedelay || element.dataset.deleteDelay || 800;
+
+  this.progress = { word: 0, char: 0, building: true, looped: 0 };
+  this.typing = true;
+
+  var colors = element.dataset.colors || "black";
+  this.colors = colors.split(",");
+  this.element.style.color = this.colors[0];
+  this.colorIndex = 0;
+
+  this.doTyping();
+};
+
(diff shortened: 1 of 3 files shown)


spencerbygraves:
Thanks @anthonydillon. It's looking good but it feels a bit robotic because it types and deletes at the same pace, is there a way we can vary it a little?

Also, the orange on blue doesn't have a great contrast value, could we try white?

Thanks.

steverydz (approved):
I find it very distracting but it does what it's supposed to and the code looks good, so :+1:

anthonydillon:
@spencerbygraves thanks. 
1. Updated the color to white. 
2. Added some randomness to the typing

danielmutis:
@spencerbygraves @anthonydillon don't you think maybe the thing should be a bit faster? I reckon people won't stay there long to see all the words we want to put in there, they may even scroll before the first word is over.

anthonydillon:
@danielmutis it was faster before but I want to give @spencerbygraves a chance to see it

danielmutis:
Okay. @spencerbygraves is typing the right animation here? I have seen this done in a similar way to how Wayne did the animation on the Juju.is commands, which I think at least allows for the whole word to be loaded at once.

spencerbygraves:
@danielmutis I'm not sure what animation you mean, the what is juju story?

I'm not tied to the typing effect, I assumed it was a request.

spencerbygraves:
The speed is better on the [jaas homepage](https://jaas.ai/), if we go for typing

spencerbygraves:
@anthonydillon thanks for the update, just a little quicker as before, or like the jaas homepage version and we're good.

Thanks.

anthonydillon:
@spencerbygraves done.

tbille:
@anthonydillon when there is a space, is it possible for you to directly jump to the next letter or add the space and the next letter together.
It looks like the writing lags when the space alone is added.

For example for `Bare metal`. Instead of adding the ` ` (between Bare and metal) could you directly add ` m`. This will avoid the weird jump.

anthonydillon:
@tbille sure thing, good idea. Done.

tbille (approved):
Looks much better!

PR #1011, opened 2021-05-17
Enable typer on small screens

anthonydillon:
## Done
Disable the small screen falls back
Enable the typer on all screen sizes

## QA
- Check out the demo on your phone and see the type in the hero
- Check large screen still works as expected


jkfran (approved):
LGTM

PR #1028, opened 2021-06-01
add badges for redis-k8s charm

cmisare-canonical:
Adding badges for the redis-k8s charm


jkfran (approved):
Thank you for doing this

PR #1079, opened 2021-07-14
Add syntax highlighting to details page overview

### LINKED ISSUE #1077: Syntax highlighting in charm READMEs
Right now, charms that have code snippets in the README (for example command line readouts), do not highlight syntax. This would probably make reading the README examples a little easier if we could support it :)

Example: https://charmhub.io/hello-juju

steverydz:
## Done
Included syntax highlighting on the details pages

## QA
- Go to https://charmhub-io-1079.demos.haus/hello-juju
- Check that the code blocks have syntax highlighting

## Issue
Fixes #1077

### DIFF
--- a/static/js/src/base/highlight.js
+++ b/static/js/src/base/highlight.js
@@ -1,7 +1,9 @@
 import hljs from "highlight.js/lib/core";
 import python from "highlight.js/lib/languages/python";
+import bash from "highlight.js/lib/languages/bash";
 import "highlight.js/styles/github.css";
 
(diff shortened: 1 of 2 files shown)


jkfran:
This is highlighting all code blocks as python, right? I don't think this is good.

I can see that we can get the language from the HTML class:
image

In this case, is bash. Could we support different kinds of languages?

steverydz:
@jkfran Thanks. This is fixed now

jkfran (approved):

PR #1082, opened 2021-07-14
Build charm resources page

steverydz:
## Done
Built the resources section of the charm details page

## QA
- Go to https://charmhub-io-1082.demos.haus/mlmd/resources/oci-image
- Go to https://charmhub-io-1082.demos.haus/postgresql/resources/wal-e
- Check that the pages look OK

[Designs](https://app.zeplin.io/project/60abb99a154f9448324fcbfe/screen/60dc831ea1e31c05f2bc33fb)

## Issue
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/2134
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/2130

## Screenshots

resources_oci

resources_file


danielmutis:
@steverydz I think it would be good to have a hover status on the base button so they can know that it's clickable. 

I will it leave it for @long-chung to review the visuals, but I think the vertical spacers in the header and the data underneath aren't the right height compared to the design.

templates/details/resources.html
```diff
@@ -16,312 +18,192 @@
...
+          {% endif %}
+        </tr>
+      </thead>
+      <tbody>
+        {% for revision in revisions[0:revisions_limit] %}
+          <tr>
+            <td>{{ revision.revision }}</td>
+            <td>-</td> <!-- update this in JS also once date field is available -->
+            {% if resource.type == "file" %}
+              <td>{{ revision.download.size * 1000 * 1000 }}</td>
```
jkfran, line 75: The API response is in bytes, it is not multiplying. But I am wondering if we can do what I have done for the size above this table. 

I used a library that automatically "humanizes" the right unit to show depending on the value.
image

Here is an example of the different outputs:
image

We can do the math by ourselves and show only MB but I think this is a better option.

@danielmutis do you think we can do this on each row and then remove the `(MB)` from the table header?

long-chung:
@steverydz Looks good. 

One minor detail; on mobile OCI image, the digest key is exposed to make it easier to view as there are no hover states on mobile. Is this achievable? 

Certainly not a blocker to release but would be great if it can be fixed.

[Link to design](https://zpl.io/2pRWy5j) 
(Click "MB" toggle at the top to switch view)

templates/details/resources.html
danielmutis: @jkfran @steverydz I think the converting the sizes would be great. I wonder if it would always stay on MB or it would go up to GB? Because if it's not consistent then we can't keep it in the header only. Otherwise we can just display all sizes in MB.

jkfran: @danielmutis We can't know how big the resource could be or even how small. IMO it is not helpful to fix the unit to MB:
image

danielmutis: @jkfran sorry I misunderstood your comment! yes, let's keep the unit in each row.

steverydz:
@danielmutis I've fixed the height of the separators and added the hover state

@long-chung I've made the digest key visible on mobile

danielmutis:
@steverydz can we style the one in the header in the same way but adjusted to the height of the text there?

jkfran (approved):
Looks good to me 👍🏽 

For the missing dates, we have [this bug](https://bugs.launchpad.net/snapstore-server/+bug/1936191) on Launchpad and looks like they are working on it.

steverydz:
@danielmutis we can, but then that breaks the pattern we use elsewhere, e.g. https://charmhub-io-1082.demos.haus/postgresql/configure

danielmutis:
@steverydz right. I imagine changing the styling everywhere would be a bigger job so let's leave as is.

PR #1085, opened 2021-07-15
Increase width of resources digest key

steverydz:
## Done
Increased the width of the resources digest key popup

## QA
- Go to https://charmhub-io-1085.demos.haus/mlmd/resources/oci-image
- Click the digest key under the title
- Check that the width of the popup is OK


jkfran (approved):
Code looks good 👍🏽

PR #1087, opened 2021-07-15
Fix tab focus on digest popup

### LINKED ISSUE #1086: Show digest key on resources page when tabbing to element
On https://charmhub.io/mlmd/resources/oci-image if you click on digest you see a popup. This doesn't work with a keyboard. It should be shown on focus.

steverydz:
## QA
- Go to https://charmhub-io-1087.demos.haus/mlmd/resources/oci-image
- Tab to the "Digest" item under the "OCI image" title
- Check that the digest key popup is visible and usable

## Issue
Fixes #1086

### DIFF
--- a/static/js/src/public/details/resources/index.js
+++ b/static/js/src/public/details/resources/index.js
@@ -1,0 +1,67 @@
+function handleTogglePanels() {
+  const panelToggleButtons = document.querySelectorAll(".js-panel-toggle");
+
+  panelToggleButtons.forEach((panelToggleButton) => {
+    panelToggleButton.addEventListener("click", (e) => {
+      const target = e.target;
+      const panel = document.getElementById(
+        target.getAttribute("aria-controls")
+      );
+
+      target.classList.toggle("is-active");
+      panel.classList.toggle("is-open");
+
+      if (panel.classList.contains("is-open")) {
+        panel.querySelector("[readonly]").focus();
+        panel.setAttribute("role", "dialog");
+      } else {
+        panel.removeAttribute("role");
+      }
+    });
+  });
+
+  document.addEventListener("keydown", (e) => {
+    const escKeyCode = 27;
+    const tabKeyCode = 9;
+
+    const openPanelToggle = document.querySelector(
+      ".js-panel-toggle.is-active"
+    );
+
+    const openPanel = document.querySelector("[readonly]");
+
+    if (e.keyCode === escKeyCode || e.keyCode === tabKeyCode) {
+      if (openPanelToggle) {
+        openPanelToggle.focus();
+        openPanelToggle.classList.remove("is-active");
+      }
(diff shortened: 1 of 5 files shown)


templates/details/resources.html
```diff
@@ -114,23 +118,25 @@ <h3 class="p-heading--5">Release history</h3>
...
-        var panel = panelToggleButton.querySelector(".p-inline-list__panel");
+        var panel = document.getElementById(
+          target.getAttribute("aria-controls")
+        );
 
-        if (target !== panel) {
-          panelToggleButton.classList.toggle("is-open");
-        }
+        target.classList.toggle("is-active");
+        panel.classList.toggle("is-open");
```
sowasred2012, line 126: Two comments:
- Only click events are being handled, so there's no way to close it without clicking; it stays open even if you tab to another part of the page. We should at least handle pressing the escape key to close the dropdown and return focus to the button.
- I think it'd be nice to focus on the input when it's revealed, since it automatically selects the value, making it even easier to copy. It highlights the whole lot in orange though, so might need a design +1.

sowasred2012 (changes requested):

templates/details/resources.html
```diff
@@ -114,23 +118,64 @@ <h3 class="p-heading--5">Release history</h3>
...
+      var escKeyCode = 27;
+      var tabKeyCode = 9;
+
+      var openPanelToggle = document.querySelector(
+        ".js-panel-toggle.is-active"
+      );
+
+      var openPanel = document.querySelector("[readonly]");
+
+      if (e.keyCode === escKeyCode) {
```
anthonydillon, line 145: Can you not just check both here? `e.keyCode === escKeyCode || e.keyCode === tabKeyCode` that would remove the duplication in the code.

steverydz: @anthonydillon yes, I'm in the process of refactoring - this was just getting it to work first

static/js/src/public/details/resources/index.js
```diff
@@ -0,0 +1,63 @@
...
+      const target = e.target;
+      const panel = document.getElementById(
+        target.getAttribute("aria-controls")
+      );
+
+      target.classList.toggle("is-active");
+      panel.classList.toggle("is-open");
+
+      if (panel.classList.contains("is-open")) {
+        panel.querySelector("[readonly]").focus();
```
sowasred2012, line 15: Last thing, and then it's good to go - could you toggle a `role="dialog"` attribute on the `panel`? At the moment, a screen reader doesn't announce anything when the dropdown appears, but it does if that attribute is added when it's shown. It reads out the entire key plus the word "selected", which is a mouthful, but I think that's better than nothing.

sowasred2012 (approved):
LGTM with one change

PR #1089, opened 2021-07-15
Update spacing and text for resource table

steverydz:
## Done
- Remove bottom margin from resources table and updated the count text
- Drive by: Fix the href of side navigation links

## QA
- Go to https://charmhub-io-1089.demos.haus/mlmd/resources/oci-image
- Check that the side nav link has the correct href value
- Check that there is no margin beneath the table and the text says "revisions" and not "versions"


sowasred2012 (approved):
LGTM

PR #1097, opened 2021-07-28
source bundle icons from charm name

### LINKED ISSUE #1096: bundle icons are sourced incorrectly
Bundle icons are pulled from the charms that make it. But instead of taking the charm name, charmhub currently uses the application name (which is an arbitrary string that charmhub should not rely on) to source the icons. Instead, charmhub should use the charm name to source the icons.

For example:

The [lma-light bundle](https://charmhub.io/lma-light) ([gh](https://github.com/canonical/lma-light-bundle)) lists the following in `bundle.yaml`:
```yaml
applications:
  alertmanager:
    charm: alertmanager-k8s
    # ...
  grafana:
    charm: grafana-k8s
    # ...
```

As a result, charmhub renders the following:
```html
  <div class="p-bundle-icons">
    <img src="/alertmanager/icon" alt="" />
    <img src="/grafana/icon" alt="" />
    
    <span class="p-bundle-icons__count u-text--muted">
      +1
    </span>
    
  </div>
```
The grafana icon show up, because ["grafana"](https://charmhub.io/grafana) happens to be an existing _different_ charm (compare to ["grafana-k8s"](https://charmhub.io/grafana-k8s)).

sed-i:
## Done
- update icon source from "application" name to charm name

## Issue / Card
Fixes #1096

### DIFF
--- a/templates/details/_details-header.html
+++ b/templates/details/_details-header.html
@@ -7,8 +7,8 @@
             {% if package.store_front.bundle.applications %}
               {% set charms = package.store_front.bundle.applications.keys() | list %}
               <div class="p-bundle-icons">
-                <img src="/{{ charms[0] }}/icon" alt="" />
-                <img src="/{{ charms[1] }}/icon" alt="" />
+                <img src="/{{ package.store_front.bundle.applications[charms[0]]["charm"] }}/icon" alt="" />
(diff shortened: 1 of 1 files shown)


jkfran (approved):

jkfran:
@sed-i thank you so much for this work!

PR #1114, opened 2021-08-20
Fix duplication of series

jkfran:
## Done
- Fix duplication of series

## How to QA
- Compare https://charmhub-io-1114.demos.haus/juju-controller with https://charmhub.io/juju-controller
- We shouldn't have duplication in the Ubuntu versions

## Issue / Card
Fixes #2179


steverydz (approved):
LGTM 👍

Issue #1073, opened 2021-07-12
CentOS identified as Ubuntu on Charmhub


heitorPB:
Charmhub shows CentOS 7 as Ubuntu 7:

![slurmd](https://user-images.githubusercontent.com/13461702/125325476-5c773000-e317-11eb-801f-9b5085b8f460.png)

The [slurmd](https://charmhub.io/slurmd?channel=edge) charm has the following [`charmcraft.yaml`](https://github.com/omnivector-solutions/slurm-charms/blob/master/charm-slurmd/charmcraft.yaml):

```yaml
type: charm
bases:
  - build-on:
      - name: ubuntu
        channel: "20.04"
    run-on:
      - name: ubuntu
        channel: "20.04"
        architectures: [amd64]
      - name: centos
        channel: "7"
        architectures: [amd64]
```

The `run-on` section for CentOS is not correctly displayed on Charmhub

heitorPB:
And we can see `centos7` with `juju info`:
```bash

$ juju info slurmd
name: slurmd
charm-id: PNY1jWPbggzG1NZvpftYqLgg86qKyNtd
summary: |
  Slurmd, the compute node daemon of Slurm.
publisher: Omnivector Solutions
supports: focal, centos7
subordinate: false
store-url: https://charmhub.io/slurmd
description: |
  This charm provides slurmd, munged, and the bindings to other utilities
  that make lifecycle operations a breeze.

  slurmd is the compute node daemon of Slurm. It monitors all tasks running
  on the compute node , accepts work (tasks), launches tasks, and kills running tasks upon request.
relations:
  provides:
    slurmd: slurmd
  requires: {}
channels: |
  latest/stable:     0.6.6  2021-07-01  (19)  4MB
  latest/candidate:  0.6.5  2021-06-28  (18)  4MB
  latest/beta:       ↑
  latest/edge:       0.6.6-1-g07b1fc0  2021-07-05  (20)  4MB
```

danielmutis:
@heitorpb thanks for reporting this issue. 

I think the correct way to show this is as in this design. 

![Screenshot 2021-08-06 at 10.21.59.png](https://images.zenhubusercontent.com/5e1c5367f4dbe2bfe4c77ad8/787e4288-1474-496b-ba8f-85a56873b06d)

heitorPB:
Hey @danielmutis, this is great!

I have a few comments about this screenshot:
[...]

PR #1178, opened 2021-11-18
Show multiple OS in header

steverydz:
## Done
Updated the charm listing header to display multiple OS in the header

## QA
- Go to https://charmhub-io-1178.demos.haus/slurmd
- Go to https://charmhub-io-1178.demos.haus/openstack-dashboard
- Check that the Operating systems under the platform heading in the header is inline with [the design](https://camo.githubusercontent.com/87916e4503b7d3b4f655e3425cd89121e46171860f18cbd97f0c1781e495c4c8/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3565316335333637663464626532626665346337376164382f32666463366536342d626665322d343534352d383862612d613866393262386635643935)

## Issue
Fixes https://github.com/canonical-web-and-design/snap-squad/issues/2229

### DIFF
--- a/templates/details/_details-header.html
+++ b/templates/details/_details-header.html
@@ -86,30 +86,36 @@
         </small>
       </p>
 
-      {% for base in channel_bases %}
+      {% if package.store_front["deployable-on"]|length == 1 and package.store_front["deployable-on"][0] == "kubernetes" %}
         <div class="series-base">
           <div class="series-base__title">
-            {% if base["name"] == "ubuntu" %}
-              <img src="https://assets.ubuntu.com/v1/adac6928-ubuntu.svg" alt="Ubuntu" width="89" height="20" class="p-image--base">
-            {% elif base["name"] == "kubernetes" %}
-              <img src="https://assets.ubuntu.com/v1/9f8a8273-2018-logo-kubernetes+%282%29.svg" alt="kubernetes" width="160" height="27" class="p-image--base">
-            {% elif base["name"] == "centos" %}
-              <img src="https://assets.ubuntu.com/v1/fd5cc5d8-CentOS.svg" alt="CentOS" width="85" height="24" class="p-image--base">
-            {% else %}
-              {{ base["name"] }}
-            {% endif %}
-          </div>
-          <div>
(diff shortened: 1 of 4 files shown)


long-chung:
@steverydz Looking good. Are you able to add the centOS logo to replace the text? 
[CentOS.svg.zip](https://github.com/canonical-web-and-design/charmhub.io/files/7564376/CentOS.svg.zip)

templates/details/_details-header.html
```diff
@@ -80,21 +80,36 @@ <h1 class="p-media-object__title">
...
-        {% endfor %}
-      </div>
-      {% endif %}
+      <p style="margin-bottom: 0.5rem;">
+        <small class="u-no-padding--top u-text--muted">
+          Platform:
+        </small>
+      </p>
+
+      {% for base in channel_bases %}
```
jkfran, line 89: We shouldn't add this new variable `channel_bases` into the template, this will work on the overview tab but not other tabs:
https://charmhub-io-1178.demos.haus/slurmd/docs

My next comment is also related to this.

steverydz:
@long-chung thanks, updated

webapp/store/views.py
```diff
@@ -159,6 +159,29 @@ def details_overview(entity_name):
...
+                    "name": i["name"],
+                    "channels": [],
+                }
+            )
+
+    for i in channel_bases:
+        for b in bases:
+            if b["name"] == i["name"]:
+                i["channels"].append(b["channel"])
+
```
jkfran, line 184: I think this is great but we should do it on `logic.py`. Take a look into the `extract_series` function on logic.py I think we need to replace this function with another one that returns a similar output to this. And then include it on `add_store_front_data` so we can use something like `package.store_front.bases` on the templates everywhere.

If we remove `extract_series` we need to make sure is compatible with the previous code everywhere.

steverydz:
@jkfran I've addressed your comments

jkfran (approved):
LGTM!

heitorPB:
That's awesome! Thanks! 
This PR closes #1073 also

PR #1195, opened 2021-12-02
Remove "Publisher suggests" for libraries

### LINKED ISSUE #1165: Library layout template: rephrase "Publisher suggests"
#447 and #509 Introduced the "publisher suggests" phrasing ~1yr ago and we might have a different understanding/perception of this now, which could potentially benefit some rephrasing.

https://github.com/canonical-web-and-design/charmhub.io/blob/e02b304805abaa6fbc15eea5513d4fdc6b6a674e/templates/details/libraries/libraries_layout.html#L29-L36

Specifically:
- Since a charm lib must be owned by a charm, there is only one charm that is the source of truth for a given library, so "publisher suggests" may seem a bit of a weak statement. Maybe "official libraries" or "original libraries" or "associated libraries" is a better phrasing, or maybe this title should be dropped entirely?
- It is common to use charm libs for relation management but the tooltip doesn't mention it explicitly.

RFC: @johnsca @mmanciop

jkfran:
## Done
- Remove "Publisher suggests" for libraries

## How to QA
- Visit the demo in your browser: /nginx-ingress-integrator/libraries/ingress

## Issue / Card
Fixes #1165

### DIFF
--- a/templates/details/libraries/libraries_layout.html
+++ b/templates/details/libraries/libraries_layout.html
@@ -24,20 +24,6 @@
           </ul>
           {% if libraries %}
           <ul class="p-side-navigation__list">
-            <li class="p-side-navigation__item--title">
-              <span class="p-side-navigation__text">
-                Publisher suggests
-                <div class="p-side-navigation__status u-no-margin--left u-hide--small">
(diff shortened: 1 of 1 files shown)


steverydz (approved):
LGTM 👍

PR #1202, opened 2021-12-15
Add opensearch to site

### LINKED ISSUE #1121: Add opensearch
It would be really nice to have the option to search for charms directly from the address bar like so: `ch nginx`

Screenshot from 2021-09-08 16-50-39

steverydz:
## Done
Added [opensearch](https://developer.mozilla.org/en-US/docs/Web/OpenSearch)

## QA
Can't QA until on live

## Issue
Fixes #1121

### DIFF
--- a/static/opensearch.xml
+++ b/static/opensearch.xml
@@ -1,0 +1,7 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+	<ShortName>Charmhub</ShortName>
+	<Description>Charmhub - The Open Operator Collection</Description>
+	<InputEncoding>UTF-8</InputEncoding>
+	<Image width="64" height="64" type="image/png">https://assets.ubuntu.com/v1/93dc475c-jaas-favicon.png</Image>
+	<Url type="text/html" method="get" template="https://charmhub-io/?q={searchTerms}"/>
+</OpenSearchDescription>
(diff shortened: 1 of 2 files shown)


static/opensearch.xml
```diff
@@ -0,0 +1,7 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+	<ShortName>Charmhub</ShortName>
+	<Description>Charmhub - The Open Operator Collection</Description>
+	<InputEncoding>[UTF-8]</InputEncoding>
```
mrgnr, line 4: I think the square brackets are not needed here and in the next line.

mrgnr (changes requested):
Aside from the square brackets, the code looks good to me from the documentation I've read. It might be good to have another pair of :eyes: before merging though.

static/opensearch.xml
steverydz: Done

steverydz:
@mrgnr I've made the requested change. Can you please approve if you're happy?

static/opensearch.xml
```diff
@@ -0,0 +1,7 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+	<ShortName>Charmhub</ShortName>
+	<Description>Charmhub - The Open Operator Collection</Description>
+	<InputEncoding>UTF-8</InputEncoding>
+	<Image width="16" height="16" type="image/x-icon">[https://assets.ubuntu.com/v1/5d4edefd-jaas-favicon.png]</Image>
```
anthonydillon, line 5: Does this image need to be 16x16 as its 256 at the moment. Also the square brackets here will fail?

anthonydillon (changes requested):
Looks good (interesting I didnt know know about this). Few comments about the image used.

static/opensearch.xml
anthonydillon: Square brackets are used in the example to show something to replace.

anthonydillon: Type should be "image/png" to. x-icon is for `.ico` images

steverydz: @anthonydillon I've replaced the image with 64x64 so that should make it look good on higher res screens too

mrgnr (approved):
LGTM :+1:

PR #1224, opened 2022-01-07
Add some redirects which have been promoted during the switch

anthonydillon:
## Done
Added two redirects to cover some recently promulgated charms

## How to QA
- Test the redirects


jkfran (approved):
LGTM

PR #1235, opened 2022-01-18
Update heading class names

steverydz:
## Done
Updated the names of heading classes to [the current Vanilla pattern](https://vanillaframework.io/docs/base/typography#heading-classes)

## QA
Check that the classes match the Vanilla docs


jkfran (approved):
LGTM

PR #1301, opened 2022-03-23
Update store module and support new exception

jkfran:
## Done
- Remove `StoreApiCircuitBreaker` since we are not using it anymore
- Update store module and support new exception

## How to QA
- Go to any 404 page on purpose and check that it is working.


Lukewh (approved):
:+1: LGTM

PR #1306, opened 2022-03-28
Added support for type hints

### LINKED ISSUE #1167: Unreadable methods signature in charms library with typehints
Hello,

I was checking out the library for the Loki charm in https://charmhub.io/loki-k8s/libraries/loki_push_api and saw this:
loki-lib

The source code is available at https://github.com/canonical/loki-operator/blob/main/lib/charms/loki_k8s/v0/loki_push_api.py#L349 and looks like:

```python
def _validate_relation_by_interface_and_direction(
    charm: CharmBase,
    relation_name: str,
    expected_relation_interface: str,
    expected_relation_role: RelationRole,
):
```

Maybe this happens because of the newlines and the type hints?

What I'd like to see in CharmHub was the signature to be correctly rendered with the type hints.

jkfran:
## Done
- Added support for type hints

## How to QA
- Visit the demo /loki-k8s/libraries/loki_push_api and check that it is working.

## Issue / Card
Fixes #1167

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ canonicalwebteam.discourse==4.0.8
 canonicalwebteam.image-template==1.3.1
 canonicalwebteam.search==1.0.0
 canonicalwebteam.store-api==3.2.9
-canonicalwebteam.docstring-extractor==1.0.0
+canonicalwebteam.docstring-extractor==1.1.0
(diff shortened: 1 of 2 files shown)


Lukewh (approved):
:+1: LGTM

PR #1362, opened 2022-06-15
Move topics lower down the page on mobile

### LINKED ISSUE #1285: Homepage- it is not clear that we can use filters for topics on small screens
on mobile view, filters are displayed after topic pages cards making it unclear that the user is also filtering those. maybe we should move the filters just before the topics?

Screenshot 2022-03-16 at 14 49 23

Lukewh:
## Done
- Change the order of elements on smaller screens

## How to QA
- Visit the [demo homepage](https://charmhub-io-1362.demos.haus/) and make the browser width the same as a phone
- See that topics move down the page

## Issue / Card
Fixes #1285

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -406,7 +406,7 @@ h3 {
 }
 
 #topics {
-  @media (max-width: $breakpoint-navigation-threshold) {
+  @media (max-width: $breakpoint-large - 1) {
     grid-row: 2;
   }
 
@@ -416,7 +416,7 @@ h3 {
       grid-template-columns: repeat(3, 1fr);
     }
 
-    @media (max-width: $breakpoint-large) {
+    @media (max-width: $breakpoint-large - 1) {
       .p-muted-heading {
         margin-bottom: 0;
         padding-top: 0;


AnnSoMuller:
Sorry, it should be the filters, the topics and then the charms :)

Lukewh:
> Sorry, it should be the filters, the topics and then the charms :)

Maybe we could ab test. My worry with putting topics above the charms is that topics may not be a user's priority (I hypothesis that topics aren't the reason a user visits charmhub.io). Sure we want to promote topics but if someone is looking on their phone for the first time (or even they're returning to look up a charm) 3 large topics, in my opinion, get in the way.

Maybe there could be some ux exploration - only show 1 topic, have a minimal view (just the title) or some different treatment for mobile?

For me, I expect the search box to search charms. Topics are a separate entity. We can "get away" with it on desktop because generally you can see the charms filter below the topics, on mobile you lose that context.

anasereijo:
@Lukewh I quite like the suggestion of the 'minimal card' - is that something easy to implement considering cards are a vanilla pattern?

AnnSoMuller:
I'm also ok with minimum cards, mobile version.

steverydz (approved):
LGTM 👍

AnnSoMuller:
@Lukewh nice for the light version of cards +1, however, it should still be bafter the filters ;) (and on top of Topics)

AnnSoMuller:
also it should behave the same way when performing a search :)

PR #1374, opened 2022-07-13
Fix opensearch template string

sed-i:
## Issue
Using the opensearch feature in firefox gives an error because the domain is incorrect, for example: `https://charmhub-io/?q=prometheus-k8s`.

## Done
Fix typo in opensearch template string introduced by #1202.


steverydz (approved):

steverydz:
Thanks for the PR!

PR #1397, opened 2022-09-01
Fix section heading not scrolling into view

### LINKED ISSUE #1396: Anchored link doesn't quite work with a new tab
Charmhub has an anchored link/id to each charm config option. e.g.,
https://charmhub.io/ceph-osd/configure#ephemeral-unmount

However, the rendering is different between when the anchored link is clicked after opening the page and when the anchored link is opened directly as a new tab.

[`ephemeral-unmount` on the left pane was clicked after opening the charm page]

Screenshot from 2022-09-01 12-03-32

^^^ `ephemeral-unmount` on the left pane is highlighted and the config title in the right pane is visible.

[the anchored link is directly opened as a new tab]
Screenshot from 2022-09-01 12-02-40

^^^ `ephemeral-unmount` on the left pane is not highlighted but the first option `aa-profile-mode` is highlighted. Also the config title in the right pane `ephemeral-unmount` is not visible.

Confirmed with both Chromium and Firefox on Ubuntu.

steverydz:
## Done
Fixed section headings being hidden under the sticky header when opened in a new tab

## How to QA
- Go to a charm configure page, e.g. https://charmhub-io-1397.demos.haus/ceph-osd/configure
- Open one of the links in the sidebar in a new tab
- Check that the heading for that section isn't hidden under the sticky header

## Issue / Card
Fixes #1396

### DIFF
--- a/templates/details/configure-charm.html
+++ b/templates/details/configure-charm.html
@@ -68,7 +68,7 @@
   function scrollToSection() {
     const hash = window.location.hash;
       const el = document.querySelector(hash);
-      const offset = 110;
+      const offset = 165;
       const elPosition = el.getBoundingClientRect().top;
       const offsetPosition = elPosition + window.pageYOffset - offset;
       window.scrollTo({ top: offsetPosition });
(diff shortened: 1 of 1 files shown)


anasereijo:
After loading the page, the first anchor link clicked hides the title, but after that shows.

Screenshot 2022-09-01 at 09 52 01

jkfran (approved):

jkfran:
I didn't see Ana's comment

PR #1428, opened 2022-11-21
WD-630 consume github api with token

codeEmpress1:
## Done
- Consume github api with account token
## Issue / Card
Fixes #

### DIFF
--- a/webapp/interfaces/views.py
+++ b/webapp/interfaces/views.py
@@ -1,5 +1,6 @@
 from flask import Blueprint, render_template
 from github import Github
+from os import getenv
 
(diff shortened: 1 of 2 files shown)


webapp/interfaces/views.py
```diff
@@ -14,8 +15,10 @@
     static_folder="/static",
 )
 
+GITHUB_TOKEN = getenv("GITHUB_TOKEN")
+
 # Todo create token
```
jkfran, line 20: Could you remove this comment?

jkfran (approved):

PR #1454, opened 2022-12-12
Update navigation

steverydz:
## Done
Updated the navigation

## QA
- Go to https://charmhub-io-1454.demos.haus/
- Check that the navigation looks and works as expected

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-845


AnnSoMuller:
sign in icon => other side  :sweat_smile:  

it seems also that the logo used for juju and Charmhub are not the same... so there it jumps a bit when you switch

PR #1485, opened 2023-02-09
Implement search and filter for Integrations tab

Lukewh:
## Done
- Implemented search and filter, which meant:
  - Introduce [RecoilJS](https://recoiljs.org) for state management
  - Introduce ReactQuery for query calls
  - Add a bunch of state
  - Do a bunch of refactoring

## How to QA
- Visit the demo and view all your favourite charms. Some of my favourite: Openstack Dashboard, Swift Storage, and postgresql
- Click on the integrations tab
- Filter the cards with the available filter
- Click around

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-510

## Screenshots
image


package.json
```diff
@@ -54,7 +54,9 @@
     "react-dom": "18.2.0",
     "react-dropzone": "^14.2.2",
     "react-markdown": "8.0.4",
+    "react-query": "^3.39.2",
```
steverydz, line 57: Pin version please

package.json
```diff
@@ -54,7 +54,9 @@
...
     "react-router-dom": "6.6.2",
+    "recoil": "^0.7.6",
```
steverydz, line 59: Pin version please

package.json
```diff
@@ -54,7 +54,9 @@
     "react-dom": "18.2.0",
     "react-dropzone": "^14.2.2",
```
steverydz, line 55: Not your fault, but pin version please

steverydz (approved):
LGTM 👍 

Just pin those new dependency versions and all good

anasereijo:
- [ ] the padding between the title and the _hr_ gets bigger if we select a lot of chips from the search dropdown. The _hr_ also moves up and down.  The size of the dropdown shouldn't affect the page.

https://user-images.githubusercontent.com/36884067/218431695-7b1c2eec-3265-4d8c-882d-4fd2d001224b.mov

- [ ] dropdown: Integrations should be singular as per all the other items
Screenshot 2023-02-13 at 10 21 56

- [ ] cards don't look like the[ design provided](https://zpl.io/jZrDmZ4)

- [ ] Description is missing from each relation

- [ ] interfaces links missing from the headings
final design
Screenshot 2023-02-13
how it looks like
Screenshot 2023-02-13 at 10 34 33

- [ ] can we limit the height of the cards to be always the same? maybe cut the name if too big
Screenshot 2023-02-13

- [ ] can we make the whole card clickable as per homepage?

Lukewh:
> * [ ]  Description is missing from each relation

Relations don't have descriptions, unless they're a new “canonical interface”

> * [ ]  interfaces links missing from the headings

Where would they link to? There's no page for an interface, unless it's a “canonical interface”

Lukewh:
> * [x]  cards don't look like the[ design provided](https://zpl.io/jZrDmZ4)

@anasereijo I've addressed most of the issues, however, “Recommended by publisher” doesn't currently exist and adding the libraries is an extra API call that will increase loading times and, in my opinion, probably isn't worth it right now. Wdyt?

anasereijo:
-1 lets chat tomorrow

PR #1554, opened 2023-03-23
Add new route for beta store

steverydz:
## Done
Created a new view for the beta store with a `noindex` request header

## How to QA
- Go to https://charmhub-io-1554.demos.haus/beta-store
- Check that there is a page there
- Check that there is a `x-robots-tag` request header set to `noindex`


Lukewh (approved):
Can it just be /beta not /beta-store? Charmhub doesn't have /store

PR #1565, opened 2023-03-31
Update beta store cards

steverydz:
## Done
- Updated the beta store cards

## How to QA
- Go to https://charmhub-io-1565.demos.haus/beta-store
- Check that there are cards (the title and some platforms are missing in the API which is a known issue)


chillkang (approved):
LGTM 👍

PR #1597, opened 2023-05-31
Fix no charms notication

steverydz:
## How to QA
- Go to https://charmhub-io-1597.demos.haus/interfaces/mutual_tls/draft
- Check that there is a notification and no "Charms" heading
- Go to https://charmhub-io-1597.demos.haus/interfaces/mysql_client/draft
- Check that there is **not** a notification and there **is** a "Charms" heading


chillkang (approved):
LGTM 👍

PR #1607, opened 2023-06-14
Cleanup beta package code

steverydz:
## Done
Removed most `useState` and `useEffect` hooks and get data from search params directly

## How to QA
- Go to https://charmhub-io-1607.demos.haus/beta-store
- Check that the filters work as expected

### DIFF
--- a/static/js/beta-store/components/App/App.tsx
+++ b/static/js/beta-store/components/App/App.tsx
@@ -5,7 +5,14 @@ import { QueryClient, QueryClientProvider } from "react-query";
 import Packages from "../Packages";
 
 function App() {
-  const queryClient = new QueryClient();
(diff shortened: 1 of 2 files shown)


Lukewh (approved):
:+1: LGTM

Lukewh:
There seems to be a couple of edge cases that can happen, that maybe can be addressed in separate PRs:
- when the page first loads, the filters aren't clickable, but they're also not disabled, until I click once somewhere, then I can click on elements in the page - a very strange one
- if I have filters changed / I'm on another "page" of results, if I leave the browser and come back, the items reset to the first page with no filter. I think this might be the refetch setting on useQuery

Lukewh (approved):

PR #1652, opened 2023-08-15
Create discourse search

codeEmpress1:
## Done
- Add endpoint for searching discourse
- Eefactor docs endpoint
## How to QA
- Go to docs, it should return a list of 5 topics from `doc` category and its subcategories only
- Go to docs/all, it should return a list of 50 topics from `doc` category and its subcategories only
- Go to topics, it should return a list of 5 topics from categories other than `22`
- Go to topics/all, it should return a list of 50 topics from categories other than `22`
## Issue / Card https://warthogs.atlassian.net/browse/WD-5781
Fixes #


steverydz (approved):
LGTM 👍

PR #1668, opened 2023-09-15
Fix q param being removed when using filters

steverydz:
## Done
Fixed the `q` parameter being removed from the URL when the filters are being used

## How to QA
- Go to https://charmhub-io-1668.demos.haus/?q=kafka
- Use the filters and check that the `q` parameter remains in the URL

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-6297


chillkang (approved):
LGTM 👍

PR #1674, opened 2023-09-28
Release beta store

steverydz:
## Done
Made the beta store the default homepage

## QA
- Go to https://charmhub-io-1674.demos.haus/
- Click around and make sure it works as expected


Lukewh (approved):
:+1: LGTM

AnnSoMuller:
Hi there, I have seen many things that would need some improvements, I'll make the list here and let's see what falls under this issue or (an)other(s):

1. Kubernetes filter doesn't seem to work
2. it it very slow when I use a filter
3. the rebranding made the navigation move to the left see juju.is We will need to align there
4. I don't see the charms icons on Bundles

AnnSoMuller:
Hi Steve, I still see few problems:

1. some filters don't seem to work (e.g. Loggin and Tracking)
Screenshot 2023-09-29 at 15 56 01 
vs 
Screenshot 2023-09-29 at 15 56 14
2. We are missing the number a packages available
Screenshot 2023-09-29 at 15 58 16
vs
Screenshot 2023-09-29 at 15 58 25
3. Packages visual designs are still very different in terms of structure/hierarchy, icons sizes, alignement, can improve there? or it is a separate issue?
Screenshot 2023-09-29 at 16 01 26
vs
Screenshot 2023-09-29 at 16 01 44
4. I see the placeholder with initials for bundles but not actual icons
5. Why the result order is different? Is it on purpose? is it because of a fake list for now? ^^

steverydz:
@AnnSoMuller I've addressed your comments

1. Filters are fixed
2. Number of packages is not there
3. The design of the cards is what we agreed upon a while back when we started the alignment with Ana B.
4. Bundle icons are fixed
5. The order doesn't seem different to me when applying the same filters? That's not a front-end issue as it just shows what we get from the API. I don't think it's a problem though.

AnnSoMuller:
Hi Steve,
Thanks for all those fixes!
I still have a couple of comments :)

1. some packages don't have their name displayed 
Screenshot 2023-10-05 at 10 27 40
Screenshot 2023-10-05 at 10 15 36
2. Visually, I have [this](https://app.zeplin.io/project/63cfa77260818d13bbbf1cb1/screen/63da2a4afae89a110abe8fc3) as the latest for cards.
Screenshot 2023-10-05 at 10 21 18
To be validated with Visual but it seems that
- packages names should be just bold text without being H3
- publisher in Italic
- less space bottom after description and after footer
- more condensed height
- 3 lines max (finish with "...")
- add the number left for charms i Bundles => e.g. +22

AnnSoMuller:
All good for me :)

PR #1688, opened 2023-10-12
Fix the width of the grid

steverydz:
## Done
Fixed the width of the grid to match juju.is

## How to QA
- Go to https://charmhub-io-1688.demos.haus/
- Compare site width to https://juju.is


Lukewh (approved):
:+1: LGTM

PR #1705, opened 2023-11-07
Fix duplicate platforms showing on charms

steverydz:
## Done
Fixed issue with duplicate platforms showing on some charms

## How to QA
- Go to https://charmhub-io-1705.demos.haus/ovn-chassis
- Check that there is only one platform
- Go to https://charmhub-io-1705.demos.haus/cos-proxy
- Check that there is only one platform


codeEmpress1:
LGTM!

PR #1731, opened 2024-01-08
Reset pagination on search

### LINKED ISSUE #1726: pagination not reset when updating search
When doing a search, the site sticks to the previously viewed result page number.
This can be confusing to users. Also it can result in a buggy behaviour, where no result page is active (because previous one is out of range) and a faulty message is displayed (e..g ```Showing 37 to 38 of 14 results for "logging".```)

Steps to reproduce:
1. display all charms
2. access result page 4
3. search for "logging"

Desired behaviour would be to always display result page 1 when doing a search.

steverydz:
## Done
Reset the pagination on search

## How to QA
- Go to https://charmhub-io-1731.demos.haus/
- Use the pagination to go to page 4
- Search for "logging"
- Check that the `page` query parameter has been removed
- Check that the results count above the search results is correct

## Issue / Card
Fixes #1726
Fixes https://warthogs.atlassian.net/browse/WD-8112

### DIFF
--- a/static/js/store/components/Banner/Banner.tsx
+++ b/static/js/store/components/Banner/Banner.tsx
@@ -21,6 +21,7 @@ function Banner({ searchRef, searchSummaryRef }: Props) {
               e.preventDefault();
 
               if (searchRef.current && searchRef.current.value) {
+                searchParams.delete("page");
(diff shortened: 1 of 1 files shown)


chillkang (approved):
LGTM👍

PR #1747, opened 2024-01-25
Accept and reject invites

steverydz:
## Done
Added ability to accept and reject invites

## How to QA
- In a private browser create a new account (e.g. firstname.lastname+testing@canonical.com)
- In your regular account go to https://charmhub-io-1747.demos.haus/<CHARM_NAME>/collaboration
- Invite the new account as a collaborator
- You will see the following error: "Error: There is already an invite for this email address"
- Ignore the error as it will be addressed separately
- In the notification that appears after the invite is created, copy the link address of the "Accept invite" link
- Paste it in a private browser and login as your new account
- Once redirected you'll need to paste the link in again as the redirect loses the token
- Accept invite
- Go to https://charmhub-io-1747.demos.haus/<CHARM_NAME>/listing with your new account
- If it's worked then you should be able to access that page

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-8221

PR #1750, opened 2024-01-31
Add collaborators table

steverydz:
## Done
Added the table for collaborators

## How to QA
- Go to https://charmhub-io-1750.demos.haus/<CHARM_NAME>/collaboration
- Check that there is a table of collaborators with the title of "Active shares"
- If you don't have any collaborators it's probably because you haven't accepted any invites so you should [follow the QA steps in this PR](https://github.com/canonical/charmhub.io/pull/1747).
- **Note**: Only IDs can be shown rather than names or email address because we currently don't have that data. It will be addressed once that is possible

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-8146

### DIFF
--- a/static/js/src/publisher/collaborator/components/Collaborators.tsx
+++ b/static/js/src/publisher/collaborator/components/Collaborators.tsx
@@ -12,11 +12,13 @@ import {
   Input,
 } from "@canonical/react-components";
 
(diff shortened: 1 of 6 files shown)


chillkang (approved):
LGTM 👍

PR #1753, opened 2024-02-02
feat: persist redirect params after logins

Lukewh:
## Done
- Added a two of functions for capturing parameters before login redirect and after
  - `param_redirect_capture` captures the state in a cookie if the URL signature matches. 
    - Sets cookie expiration to 10 days - this is conservative and we might consider reducing this as a default
  - `param_redirect_exec` pops the state out of the cookie (and removes the cookie) if the path matches.
- Added a call to `param_redirect_capture` from the `login_required` decorator
- Added a new decorator `cached_redirect` that calls `param_redirect_exec` and uses the current URL path to match a cookie

## How to QA
**With an invitation URL to a charm**
- In a private browser that you're not logged in:
  - Visit the URL
  - When redirected to login.ubuntu.com, open the demo in a different tab and confirm a `param_redirect` cookie has been created for the demo URL.
  - Go through login process
  - after login you should be redirected back to the original URL with parameters
  - The previously created cookie should also be removed

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-8484

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Install python dependencies
         run: |
           python3 -m pip install --upgrade pip
-          sudo pip3 install flake8 black
+          sudo pip3 install -r requirements.txt
 
       - name: Lint python
         run: yarn lint-python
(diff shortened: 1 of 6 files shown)


steverydz (approved):
LGTM 👍

PR #1763, opened 2024-02-14
fix the side-navigation on resources tab where the is-active class is being applied on all <a> elements

chillkang:
## Done
- fix the side-navigation on resources tab where the `is-active` class is being applied on all `<a>` elements

## How to QA
- go to 
- check if you can select only one side-navigation link at a time 

## Issue / Card
Fixes #https://warthogs.atlassian.net/browse/WD-8885


steverydz (approved):
LGTM 👍

PR #1768, opened 2024-02-21
add tests for interfaces view

codeEmpress1:
## Done
- Add tests for interfaces view
## Issue / Card
Fixes #


tests/interfaces/test_single_interface.py
```diff
@@ -0,0 +1,103 @@
+import json
+from unittest.mock import patch
+from unittest import TestCase
+
+import responses
+from webapp.app import app
+from webapp.integrations.logic import Interfaces
+
+mock_interface_yaml = """
```
abbiesims, line 9: This isn't used anywhere, has it been setup for future tests or can we remove it?

tests/interfaces/test_single_interface.py
```diff
@@ -0,0 +1,103 @@
...
+class TestSingleInterface(TestCase):
+
+    def setUp(self):
+        app.config["TESTING"] = True
+        app.config["DEBUG"] = True
+        self.client = app.test_client()
+        self.test_interfaces = Interfaces()
+        # mock_github_client = patch("webapp.integrations.logic.github_client")
+
+        self.github_interfaces_url = "".join(
```
abbiesims, line 35: Is there a reason you are using `.join` instead of creating a variable with one string?

tests/interfaces/test_single_interface.py
codeEmpress1: I can take this out for now.

tests/interfaces/test_single_interface.py
codeEmpress1: The string is longer that the line limit of 79 so I had to break into two lines and join

tests/interfaces/test_single_interface.py
abbiesims: Did you want to remove this in this revision? Not sure if the changes have gone through

codeEmpress1: Yes, it's done now

abbiesims (approved):
LGTM! :+1:

PR #1773, opened 2024-02-27
Add inks for integrations

chillkang:
## Done
- Ensure `/interfaces/:interface` does not 404 on production
- Add the uuid package and install TypeScript type definitions for uuid
- Include links for interfaces on the integration tab
- Create a custom component to render a mermaid diagram and use it with React Markdown

## How to QA
- Go to https://charmhub-io-1773.demos.haus/kafka/integrations
- Check if you can click on the second part of a title, e.g., `kafka_client` on `kafka-client | kafka_client`
- Check if `See all interfaces` link is hidden in the header
- Check if you can see `Developer Documentation` section rendered including a mermaid diagram

## Issue / Card
Fixes #https://warthogs.atlassian.net/browse/WD-8668

## Screenshot

1. Integrations page
Screenshot 2024-02-29 at 6 43 42 AM

2. Interface Detail page, e.g., kafka_client
Screenshot 2024-03-25 at 7 44 17 AM

### DIFF
--- a/static/js/src/interfaces/components/InterfaceDetails/InterfaceDetails.tsx
+++ b/static/js/src/interfaces/components/InterfaceDetails/InterfaceDetails.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 import { Link, useParams } from "react-router-dom";
 import { useQuery } from "react-query";
 import { formatDistance } from "date-fns";
@@ -42,7 +42,6 @@ type Props = {
 };
 
 function InterfaceDetails({ interfaceItem }: Props) {
-  const [showLink, setShowLink] = useState(false);
   const { interfaceName, interfaceStatus } = useParams();
   const shouldFetchData = () => {
     if (interfaceItem && interfaceItem.name === interfaceName) {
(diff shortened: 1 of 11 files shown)


static/js/src/interfaces/components/InterfaceDetails/InterfaceDetails.tsx
```diff
@@ -89,9 +90,11 @@ function InterfaceDetails({ interfaceItem }: Props) {
           {interfaceData?.name && <>{interfaceData?.name}</>}
           {!interfaceData?.name && interfaceName}
         </h1>
-        <p>
-          <Link to="/interfaces">See all interfaces</Link>
-        </p>
+          {showLink && (
```
steverydz, line 93: Is there a case when `showLink` wouldn't be `false`? `setShowLink` is never called.

steverydz (approved):
LGTM :+1:

Just one question in the code

static/js/src/interfaces/components/InterfaceDetails/InterfaceDetails.tsx
chillkang: No, because we want to hide the link for the time being. And I agree that the way `useState `is used isn't ideal in this scenario. Then would it be okay if I remove the link itself?

steverydz: Yes, I think if the link isn't needed right now it could be removed, unless we know when we're going to use it.

PR #1784, opened 2024-04-03
fix: accept navigation items without children

skatsaounis:
## Done

Add check for length of children of navigation items. Make sure that the length is greater than 1 before trying to access the second item of the list

## How to QA

- Access: http://0.0.0.0:8045/maas-region and get the error when on main branch
- Access: http://0.0.0.0:8045/maas-region and get the page when on PR branch

## Issue / Card
Fixes #

## Screenshots

Without the fixes:

image

With the fixes:

image


Lukewh:
Demo: https://charmhub-io-1784.demos.haus/

chillkang (approved):
LGTM👍

PR #1790, opened 2024-04-15
Remove block from creating expired or revoked invites

steverydz:
## Done
Only validate unique invites if they are pending

## How to QA
- Go to https://charmhub-io-1790.demos.haus/<CHARM_NAME>/collaboration
- Click the "Add new collaborator" button
- Type in an email address for an invite that is either revoked or expired
- Check that you don't get a validation error
- Type in an email address for an invite that is already pending
- Check that you do get a validation error


chillkang (approved):
LGTM👍

PR #1815, opened 2024-05-03
Add Sentry to publisher app

steverydz:
## Done
Added Sentry to publisher React app

## How to QA
Can't QA until live. I've added a function which triggers an error when called from the console so I can see if it makes it to Sentry. Then it will be removed in another PR.

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-10980


Lukewh (approved):
:+1: LGTM

PR #1816, opened 2024-05-03
Make sure Sentry DSN is accessible from the front-end

steverydz:
## Done
Made Sentry DSN available on the front-end

## How to QA
- Go to https://charmhub-io-1816.demos.haus/<CHARM_NAME>/listing
- Type `window.SENTRY_DSN` and check that it has a value


Lukewh (approved):
:+1: LGTM

PR #1831, opened 2024-05-24
set coverage threshold

codeEmpress1:
## Done
- set threshold for test coverage
## Testing
- [ ] This PR has tests
- [x] No testing required (explain why):

## Issue / Card https://warthogs.atlassian.net/browse/WD-11546
Fixes #

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -15,11 +15,11 @@ coverage:
     project:
       javascript:
         target: auto
-        threshold: 0%
+        threshold: 20%
         flags:
           - javascript
       python:
         target: auto
-        threshold: 0%
+        threshold: 44%
         flags:
           - python


codecov.yml
```diff
@@ -15,11 +15,11 @@ coverage:
     project:
       javascript:
         target: auto
-        threshold: 0%
+        threshold: 20%
```
abbiesims, line 18: Where are these numbers coming from? In the doc, it says backend has a current coverage of 59% and frontend has 2.8%

codeEmpress1: Thanks for pointing this out. I would say this is a calculated assumption because the report from TICS does not truly represent the state of our codebase at the moment. A lot of the coverage reports error which calculates it at 0%. Though I could take this to the 2.8% that was reported but I was planning to update the coverage as soon as Luke re-runs the test and we have a reliable figure.

abbiesims: Okay, thanks for explaining, makes sense!

abbiesims (approved):
LGTM :+1:

Issue #1793, opened 2024-04-15
Charmhub WEB shows (amd64?) architecture constantly hiding arm64 charm revisions


taurus-forever:
Hi,

Compare ARM and AMD revisions for [grafana-agent charm](https://charmhub.io/grafana-agent?channel=latest/edge):

![Screenshot from 2024-04-15 18-05-12](https://github.com/canonical/charmhub.io/assets/1928266/62b43b1f-afda-4efc-a2f6-44e7ca9be1a2) ![Screenshot from 2024-04-15 18-05-16](https://github.com/canonical/charmhub.io/assets/1928266/d9cea249-e284-4e21-b47b-b34fd0e65aed)

In fact revisions are:

```
Track    Base                  Channel    Version    Revision                                                                               
latest   ubuntu 20.04 (amd64)  stable     66         66                                                                                     
                               candidate  95         95                                                                                     
                               beta       95         95                                                                                     
                               edge       95         95                                                                                     
         ubuntu 20.04 (arm64)  stable     67         67                                                                                     
                               candidate  67         67                                                                                     
                               beta       67         67                                                                                     
                               edge       93         93                                                                                     
         ubuntu 22.04 (amd64)  stable     65         65                                                                                     
                               candidate  65         65                                                                                     
                               beta       65         65                                                                                     
                               edge       96         96                                                                                     
         ubuntu 22.04 (arm64)  stable     68         68                                                                                     
                               candidate  68         68                                                                                     
                               beta       68         68                                                                                     
                               edge       94         94                   
```

It caused wrong deployments and unnecessary [bugreports](https://github.com/canonical/grafana-agent-operator/issues/89).

Tnx for the fix here!

taurus-forever:
BTW, architecture filter works well for https://charmhub.io/s3-integrator and do not show arm at all for https://charmhub.io/postgresql but supported (on default track `14`):
```
14       ubuntu 22.04 (amd64)  stable                  363                            363         -                                                                                                                                                   
                               candidate               363                            363         -                                                                                                                                                   
                               beta                    368                            368         -                                                                                                                                                   
                               edge                    396                            396         -                                                                                                                                                   
         ubuntu 22.04 (arm64)  stable                  -                              -           -                                                                                                                                                   
                               candidate               -                              -           -                                                                                                                                                   
                               beta                    -                              -           -                                                                                                                                                   
                               edge                    395                            395         -                                                                                                                                                   
```

roosterfish:
[...]

Issue #1741, opened 2024-01-19
List of Ubuntu Series (Bases) is incomplete for charms with multiple binary builds (1 build per base+architecture combination)


lathiat:
When a charm has individual binary builds for multiple Ubuntu series (bases) and architectures, the list of architectures is not complete/correct.

Example case: cinder-purestorage
https://charmhub.io/cinder-purestorage

The latest release to yoga/stable has 8 builds for Ubuntu 20.04 and 22.04 on each of the architectures amd64, as you can see from juju info. However the charmhub.io page only shows Ubuntu 20.04:

```
juju info cinder-purestorage --format json | jq -r '.channels.yoga.stable[] | [.version, .revision, .architectures[0], .bases[0].channel] | @tsv'
1fabc7b	245	arm64	20.04
1fabc7b	244	arm64	22.04
1fabc7b	243	ppc64el	20.04
1fabc7b	242	ppc64el	22.04
1fabc7b	241	s390x	20.04
1fabc7b	240	s390x	22.04
1fabc7b	239	amd64	22.04
1fabc7b	238	amd64	20.04
```

Base/Architecture independent charms don't have this issue. But those also return a single entry in .channels.yoga.stable[], instead of multiple entries with the same 'version'.

We seem to have taken multiple stabs at the same or similar issue in the past so I guess we need to make sure we think through the entire problem space when solving this. Here is my attempt to do that for the cases I know about, but this knowledge may be incomplete

- A list of architectures is noted in charmcraft.yaml (build-on/run-on) but also listed in metadata.yaml. It's not clear when the metadata.yaml version is used (if ever).
- Charms may have a single build but a specific list of build-on/run-on, in which case .channels.yoga.stable[] would return a single item with multiple architectures and bases. An example is nova-compute's yoga channel. It has 20.04+22.04 with amd64+arm64+ppc64el+s390x explicitly specified but all part of a single revision/build. This currently displays correclty.
- Charms may have one build per-architecture, that supports multiple bases. An example is octavia's zed channel which has 4 builds for each architecture but each build supports both Ubuntu 22.04+22.10. This currently displays correctly.
[...]

PR #1839, opened 2024-06-03
fix: channel dropdown logic for multi-base channels

M7mdisk:
## Done

- Add "Base version" dropdown  in channels list
- Refactor the logic to show all revisions, not just latests ones.
- Show all bases associated with a channel regardless of which revision they're associated with.

## How to QA
- Go to any charm, the channel list menu should show the most recent releases for each channel, closely matching the output of `juju info charm-name`.
- When selecting an architecture, the output should the most recent channels supporting that architecture, matching `juju info charm-name --arch=arm64` for example.
- When selecting a base, the output should show the most recent channels with that base. matching  `juju info charm-name --base=ubuntu@20.04` for example.
- a combination of arch and base should work as well.
- Some good charms to test on (have many architectures and bases supported): `mysql-router`, `landscape-server`, `lxd`, `pgbouncer`.

Note: small caveat - The bases for each channel reflect all bases supported by the channel, not just the bases supported by the displayed revision

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Fixes existing logic

## Issue / Card
Fixes [WD-10284](https://warthogs.atlassian.net/browse/WD-10284),  #1793 and #1741

[WD-10284]: https://warthogs.atlassian.net/browse/WD-10284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lukasSerelis:
Still some funkiness happening. For example, the [ceph charm](https://charmhub-io-1839.demos.haus/ceph-fs), if you select s390x as architecture, a bunch of base versions pop up, which is odd, we should have those exposed earlier, but might be `juju info` issue too.

lukasSerelis:
Smaller detail, architecture dropdown order of elements seems to have changed their order. I'm not sure how it was sorted before, but would be best to sort it logically, like a-z.

lukasSerelis:
A bigger detail - we're still not displaying all bases supported. [This charm](https://charmhub-io-1839.demos.haus/pgbouncer?channel=1/stable), doesn't show ubuntu@22.04 supported at all, but if you `juju info pgbouncer --base=ubuntu@22.04` you can see that it is supported 
image

Additionally, even the charm track list show the 22.04 version for some tracks, but doesn't surface it for filtering.

lukasSerelis:
Another issue - might be out of scope for this, is that base versions aren't consistently displayed under platforms. 
image
assets/152963078/9b61d3b9-2fbb-47df-821c-9d0a41d289f4)
sometimes some versions are show, sometimes only 1

Lukewh (approved):
:+1: LGTM

PR #1848, opened 2024-06-10
fix: integrations tab shows correct charms for each channel

abbiesims:
## Done
- Changes path to query for integrations to take the selected channel into account.
- Also removes duplicated charm cards in integrations.

## How to QA
- Check https://charmhub-io-1848.demos.haus/mongodb/integrations and change channel
- Integrations supported should change depending on the channel
- Check duplicated cards against the current version (https://charmhub.io/mongodb/integrations)

## Testing
- [ ] This PR has tests
- [X] No testing required (explain why): bug fix

## Issue / Card
Fixes https://warthogs.atlassian.net/browse/WD-11698

### DIFF
--- a/static/js/src/public/details/integrations/components/App/App.tsx
+++ b/static/js/src/public/details/integrations/components/App/App.tsx
@@ -12,7 +12,14 @@ import { useRecoilState, useRecoilValue } from "recoil";
 import { filterChipsSelector, filterState } from "../../state";
 
 const getIntegrations = async (charm: string): Promise<IInterfaceData[]> => {
-  const resp = await fetch(`/${charm}/integrations.json`);
+  let resp;
+  const url = new URL(document.location.href);
+  const selectedChannel = url.searchParams.get("channel");
+  if (selectedChannel) {
+    resp = await fetch(`/${charm}/integrations.json?channel=${selectedChannel}`);
+  } else {
+    resp = await fetch(`/${charm}/integrations.json`);
+  }
   const json = await resp.json();
   if (!json.grouped_relations) {
     return [];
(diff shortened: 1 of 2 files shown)


static/js/src/public/details/integrations/components/App/App.tsx
```diff
@@ -12,7 +12,13 @@ import { useRecoilState, useRecoilValue } from "recoil";
 import { filterChipsSelector, filterState } from "../../state";
 
 const getIntegrations = async (charm: string): Promise<IInterfaceData[]> => {
-  const resp = await fetch(`/${charm}/integrations.json`);
+  let resp;
+  const selectedChannel = window.location.search.split("channel=")[1];
```
Lukewh, line 16: It's more text, but it might be more reliable to:
```
const url = new URL(document.location);
const selectedChannel = url.searchParams.get("channel");
```
Generally doing splits etc. on urls will at some point be sad.

static/js/src/public/details/integrations/components/InterfaceItem.tsx
```diff
@@ -161,15 +161,19 @@ export const InterfaceItem = ({
...
-            {charms.map((charm: ICharm) => (
-              <Col
-              size={3}
-              style={{ marginBottom: "1.5rem" }}
-              key={charm.package["display_name"]}
-              >
-                <IntegrationCard data={charm} />
-              </Col>
-            ))}
+            {charms
```
Lukewh, line 164: Spicy - how are we getting duplicates that then need filtering out? Does the API return duplicates? If the API doesn't we should probably fix the root cause in our code, rather than masking it with a filter.
If it is because we get duplicates from the API, it would be useful to have a comment here explaining why we have this and letting the store team know there are duplicates.

abbiesims: I've been digging to this, and I think `<IntegrationCard>` might be doing some of its own state management or manipulation under the hood. When I log out `charms` right before this, there is only one of each kind, but when rendering using `<IntegrationCard>`, there are duplicates :thinking:

Lukewh (approved):
:+1: LGTM

PR #1851, opened 2024-06-11
fix: add a chip to integration tabs when required property is set to true in metadata.yaml

chillkang:
## Done
- When `required` property in `metadata.yaml` is set to `true`, the `Required` chip will be added to relevant side-nav items and endpoint headings.

## How to QA
- Currently, we don't have `required` property in metadata.yaml but this will be the standard.

## Testing
- [x] This PR has tests

## Issue / Card
Fixes #https://warthogs.atlassian.net/browse/WD-9928

## Screenshots
Screenshot 2024-06-14 at 9 32 09 PM

### DIFF
--- a/static/js/src/public/details/integrations/types.ts
+++ b/static/js/src/public/details/integrations/types.ts
@@ -3,6 +3,7 @@ export interface IInterfaceData {
   interface: string;
   description: string;
   type?: "provides" | "requires";
+  optional: string;
 }
 
 export interface IFilterChip {
--- a/static/js/src/public/details/integrations/types.ts
+++ b/static/js/src/public/details/integrations/types.ts
@@ -3,7 +3,7 @@ export interface IInterfaceData {
   interface: string;
   description: string;
   type?: "provides" | "requires";
-  optional: string;
+  optional?: "true" | "false";
 }
 
 export interface IFilterChip {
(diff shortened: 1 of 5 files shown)


Lukewh:
We seem to have lost the space between `certificates` and `endpoint`?

image

chillkang:
> * https://charmhub-io-1851.demos.haus/postgresql-k8s/integrations

@Lukewh We should have the space back now!

static/js/src/public/details/integrations/types.ts
```diff
@@ -3,6 +3,7 @@ export interface IInterfaceData {
   interface: string;
   description: string;
   type?: "provides" | "requires";
+  optional: string;
```
Lukewh, line 6: Should optional be optional? It might not be in the data at all, right? So this should potentially be `optional?:` also is it a string or is it a boolean? If it's a string, we should know what the values can be so you can do something similar to the `type?: "provides" | "requires"` on the line above.

lukasSerelis:
Current solution is to pretend there is a required=true field in metadata.yaml or charmcraft.yaml and check for that. As we make it a standard it will be displayed in Charmhub.

static/js/src/public/details/integrations/types.ts
chillkang: From the data we are getting from the `["store_front"]` key, the `optional` field is implemented as a boolean value. I made changes accordingly and updated the field to `required` as this will be the standard, as Lukas mentioned.

Lukewh (approved):
:+1: LGTM

lukasSerelis:
Looks good!

PR #1855, opened 2024-06-14
Replace vm with machine

M7mdisk:
## Done
- Replace all instances of "VM" with machine, including badges
## How to QA
- Check we don't use "VM" anywhere that we shouldn't

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): copy change


Lukewh (approved):
:+1: LGTM

PR #1869, opened 2024-07-09
fix: Fix isolated module compiler error

abbiesims:
## Done
- Adds empty export to fix `--isolatedModules` error in TICS

## How to QA
Checks should pass

## Testing
- [ ] This PR has tests
- [X] No testing required (explain why): no behavioural change

### DIFF
--- a/static/js/src/base/navigation.ts
+++ b/static/js/src/base/navigation.ts
@@ -80,3 +80,5 @@ if (navAccountContainer) {
 }
 
 initNavDropdowns(".p-navigation__item--dropdown-toggle");
+
+export {};


static/js/src/base/navigation.ts
```diff
@@ -80,3 +80,5 @@ if (navAccountContainer) {
 }
 
 initNavDropdowns(".p-navigation__item--dropdown-toggle");
+
+export {};
```
Lukewh, line 84: This seems like an antipattern to satisfy the automated check-gods. What's the actual error?

Lukewh: hmm, well it's suggesting a blank export and honestly I don't think it's worth modifying the contents of the file too much for risk of breaking things so I'll :+1: this :shrug:

Lukewh (approved):

PR #1872, opened 2024-07-09
fix: Fix incorrect path names in HTML templates

abbiesims:
## Done
Fixes incorrect path names from previous compiler error PRs
- switches back to `js` from `ts`


Lukewh (approved):
:+1: LGTM

PR #1873, opened 2024-07-09
fix: Tics security fix

Lukewh:
## Done
- Added more strict checks to internal redirect

## How to QA
- Visit any charm and uppercase the charm name in the url
- Be redirected to the lowercase version
- WON'T WORK ON DEMOS you will get a 404 (which is also expected)

## Testing
- [x] This PR has tests

## Issue / Card
Fixes #


steverydz (approved):
LGTM 👍

PR #1911, opened 2024-07-26
chore: deletes unused typer.ts file

abbiesims:
## Done
Deletes unused `typer.ts` file from `static/js/src/libs`

## How to QA
All tests should pass and dotrun should run without failures

## Testing
- [ ] This PR has tests


steverydz (approved):
LGTM 👍

PR #1942, opened 2024-08-05
feat: add track form

M7mdisk:
## Done
- Create add track form to be shown when guardrails exist
- Add optional track creation params to API endpoint call

NOTE: Not in this PR:
- Empty state for newly created tracks

## How to QA
- Go to a charm with guardrails, click "Add track"
- fill add track form
- create track
## Testing
- [x] This PR has tests

## Issue / Card
Fixes WD-13661

### DIFF
--- a/static/js/src/publisher-admin/pages/Releases/AddTrackPanel.tsx
+++ b/static/js/src/publisher-admin/pages/Releases/AddTrackPanel.tsx
@@ -14,12 +14,14 @@ type AddTrackPanelProps = {
   charmName: string;
   onClose: () => void;
   setSelectedTrack: (track: string) => void;
+  onSuccess?: () => void;
 };
 
 export default function AddTrackPanel({
   charmName,
   onClose,
   setSelectedTrack,
+  onSuccess,
 }: AddTrackPanelProps) {
   const [trackName, setTrackName] = useState("");
   const [versionPattern, setVersionPattern] = useState("");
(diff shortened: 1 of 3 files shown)


static/js/src/publisher-admin/pages/Releases/AddTrackPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const response = await fetch(`/${charmName}/create-track`, {
+      method: "POST",
+      body: formData,
+    });
+
+    const responseData = await response.json();
+
+    if (response.ok) {
+      refetch();
+      setSelectedTrack(trackName);
```
abbiesims, line 56: On successful addition of a new track, can you close the side panel, switch to the new track and also show a success message like we do in Snapcraft.io? This should make it very clear to the user that the track has been created successfully

static/js/src/publisher-admin/pages/Releases/AddTrackPanel.tsx
```diff
@@ -0,0 +1,178 @@
...
+            appearance="positive"
+            onClick={submit}
+            hasIcon={loading}
+            disabled={
+              loading ||
+              !trackName ||
+              !!Object.keys(errors).filter((key) => errors[key]).length
+            }
+          >
+            {loading ? <Spinner text="Add track" /> : "Add Track"}
```
abbiesims, line 172: When you set loading, can you make the button say "Loading..." instead of keeping it as "Add track" with the loading state? Again, I'm pretty sure we do this on Snapcraft as well so would be good for consistency

abbiesims (changes requested):

abbiesims (approved):
Awesome, looks good! :+1:

PR #1960, opened 2024-08-07
feat: charming charmhub

M7mdisk:
## Done
- Create a rock for flask app
- create a charm using PaaS app charmer
- add workflow to automatically build and publish rock & charm

> [!TIP]
> Dear reviewer, some important points:
> -  make sure the files listed in `rockcraft.yaml` aren't missing anything, only those will be included in the image
> -  workflow files could use another pair of eyes. especially the rock one

## QA
- This is currently deployed on staging.charmhub.io. Make sure we have 1-1 parity with current store deployment and everything works

## Issue / Card
Fixes WD-13668

### DIFF
--- a/charm/.gitignore
+++ b/charm/.gitignore
@@ -7,3 +7,4 @@ __pycache__/
 *.py[cod]
 .idea
 .vscode/
+lib/


charm/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Local development
+
+To work on this charm locally, you first need to set up an environment, follow [this section](https://juju.is/docs/sdk/write-your-first-kubernetes-charm-for-a-flask-app#heading--set-things-up) of the tutorial.
+
+Then, you can run the following command to pack and upload the rock:
+
+```bash
+rockcraft pack
+rockcraft.skopeo --insecure-policy copy --dest-tls-verify=false oci-archive:charmhub-io*.rock docker://localhost:32000/charmhub-io:1
```
Lukewh, line 13: for some engineers this will be the first time they interact with rocks - can you add a little explanation of these commands below (or links to the appropriate docs) thanks!

charm/requirements.txt
```diff
@@ -0,0 +1 @@
+paas-app-charmer==1.*
```
Lukewh, line 1: should we pin this to a specific version? Is the paas-app-charmer stable enough to have the wildcard?

charm/src/charm.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+# Copyright 2024 Moe Isk
```
Lukewh, line 2: I'll call you Moe from now on

Lukewh: also this should probably be "Canonical" :shrug:?

Lukewh (commented):
A couple of questions / comments

.github/workflows/publish-rock.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Build assets
+        run: |
+          yarn install
+          yarn run build
+
+      - name: Setup LXD
+        uses: canonical/setup-lxd@main
+
+      - name: Setup rockcraft
+        run: sudo snap install rockcraft --classic --channel=latest/edge
```
lorumic, line 27: Forwarding [this](https://github.com/canonical/anbox-cloud.io/pull/352#discussion_r1732410572) question from @edlerd - what should we use? `stable` or `edge`? I see that kernel factory uses [`stable`](https://github.com/canonical/kernelfactory.canonical.com/blob/main/.github/workflows/deploy.yaml#L65) (CC @jw-can), but `dqlite.io` uses [`edge`](https://github.com/canonical/dqlite.io/blob/main/.github/workflows/deploy.yaml#L53). :confused: 

Same question applies to the `charmcraft` snap a few lines above.

jw-can: KF had been using `edge` until very recently, since that was the only place where the flask extensions were available. However, the extensions have been merged some while now, so you could also go ahead with using `stable`, which imho would make sense.

Small warning though: There's an issue in current rockcraft in stable, which makes [packing the rock fail](https://github.com/canonical/kernelfactory.canonical.com/actions/runs/10573677661/job/29293631716)). The suggestion is to use an older version of rockcraft.

lorumic: Thanks both for your feedback! For now, it seems safe to keep using the `edge` channel then. Marking as resolved.

Lukewh (approved):
:+1: LGTM

PR #1995, opened 2024-09-06
fix: remove unused publisher code

M7mdisk:
## Done
- Remove access to non existing file which was causing JS errors when accessing /charms
## How to QA
- go to https://charmhub-io-1995.demos.haus/charms get no console errors about non existing files or `charmhub.publisher is not defined`

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): is bug fix


steverydz (approved):
LGTM 👍

PR #2012, opened 2024-10-11
chore: Remove grid override

steverydz:
## Done
Removed a grid override that uses Vanilla variables which will change in the next release, therefore causing build errors. These classes are defined [elsewhere in the codebase](https://github.com/canonical/charmhub.io/blob/main/static/sass/_pattern_grid.scss) so removing this doesn't change anything.

## How to QA
- Go to https://charmhub-io-2012.demos.haus/data-platform-libs/libraries/data_interfaces
- Check that it looks the same as production: https://charmhub.io/data-platform-libs/libraries/data_interfaces

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): Cleanup

### DIFF
--- a/static/sass/_pattern_p-docstring.scss
+++ b/static/sass/_pattern_p-docstring.scss
@@ -134,19 +134,6 @@
   }
 
   .p-docstring {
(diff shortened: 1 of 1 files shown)


chillkang (approved):
LGTM👍

PR #2026, opened 2024-11-08
fix: only show bases that actually support the selected arch

### LINKED ISSUE #2025: The bases tags are misleading... or mabe the filter is wrong
URL: https://charmhub.io/grafana-agent-k8s

If I filter by `arm64` and `22.04`, the results seem to suggest that this very charm versions also support `20.04`

Screenshot 2024-11-08 at 18 05 30

But if I filter by `arm64` and `20.04`, there are no results, so maybe `20.04` is not supported after all (for arm specifically)?

Screenshot 2024-11-08 at 18 05 44

Filtering for amd64 is also weird, but less illustrative, only the versions are different.

M7mdisk:
## Done
- Only show bases that actually support the selected architecture
- If the filter is "any", show all bases supported across all architectures
- runby: increase "+X" threshold to 9
## How to QA
- Go to grafana-agent-k8s charm
- only see bases that are actually supported by the channel
- make sure things aren't broken for other charms
## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): only presentation change

## Issue / Card
Fixes #2025 and WD-16693

### DIFF
--- a/templates/details/_details-header.html
+++ b/templates/details/_details-header.html
@@ -104,9 +104,9 @@
             </div>
             <div>
               {% for base in bases | sort(reverse=True) %}
-                {% if loop.index <= 8 %}
+                {% if loop.index <= 9 %}
                   <span class="series-tag">{{ base }}</span>
-                {% elif loop.index == 9 %}
+                {% elif loop.index == 10 %}
                   <span class="series-tag">+{{ loop.revindex }}</span>
                 {% endif %}
               {% endfor %}


lukasSerelis:
This still doesn't solve the issue. Ideally when selecting arm64 arch, it shouldn't show the ubuntu@20.04 chip at all.

Selected arm64, shouldn't show unsupported bases at all in the table. Example of desired state added:
image

abbiesims (approved):
LGTM :+1:

codeEmpress1 (approved):

PR #2082, opened 2025-02-13
chore: remove application layout on publicise section

abbiesims:
## Done
- Removes application layout in `Publicise` tab of a package as it was redundant and causing layout issues

## How to QA
- Go to the Publicise tab on a charm/bundle that you are a collaborator on
  - e.g. https://charmhub-io-2082.demos.haus/landscape-keystone/publicise
- Check all content is visible on all sections (buttons, badges and embedded cards)
  - also check on different screen sizes (at the moment, content is getting cut off on my 13" laptop on Chrome)

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes [WD-19293](https://warthogs.atlassian.net/browse/WD-19293)

[WD-19293]: https://warthogs.atlassian.net/browse/WD-19293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/src/publisher-admin/pages/Collaboration/Collaboration.tsx
+++ b/static/js/src/publisher-admin/pages/Collaboration/Collaboration.tsx
@@ -90,74 +90,75 @@ function Collaboration() {
   return (
     <>
       <div className="l-application">
-        {showRevokeSuccess && (
-          <Notification
-            severity="positive"
-            onDismiss={() => {
-              setShowRevokeSuccess(false);
-            }}
-          >
-            The invite for <strong>{activeInviteEmail}</strong> has been
-            revoked.
-          </Notification>
-        )}
-
-        {showRevokeError && (
-          <Notification
-            severity="negative"
-            onDismiss={() => {
-              setShowRevokeError(false);
-            }}
-          >
-            There was a problem revoking the invite for{" "}
-            <strong>{activeInviteEmail}</strong>.
-          </Notification>
-        )}
-
-        {showInviteSuccess && (
-          <Notification
-            severity="positive"
-            title="An invite has been created"
-            onDismiss={() => {
-              setShowInviteSuccess(false);
-            }}
-          >
-            <p>
-              <a target="_blank" href={inviteEmailLink} rel="noreferrer">
-                Send the invite by email
-              </a>{" "}
-              or copy link:
-            </p>
-            <div>
-              <input
-                className="u-no-margin--bottom"
-                type="text"
-                readOnly
-                value={inviteLink}
-                style={{
-                  color: "inherit",
-                }}
-                onFocus={(e) => {
-                  e.target.select();
-                }}
-              />
-            </div>
-          </Notification>
-        )}
-
-        {showInviteError && (
-          <Notification
-            severity="negative"
-            onDismiss={() => {
-              setShowInviteError(false);
-            }}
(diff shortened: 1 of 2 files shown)


static/js/src/publisher-admin/routes/root.tsx
```diff
@@ -22,17 +22,9 @@ function Root() {
...
-          <div className="p-panel">
-            <div className="p-panel__content">
-              <div className="u-fixed-width">
-                <SectionNav />
-                <Outlet />
-              </div>
-            </div>
-          </div>
-        </div>
+      <div className="u-fixed-width">
```
bartaz, line 25: So this is a root of all the views? Not just the Publicise view?

Could this be a case that application layout is used for something somewhere (on other views)?
It is just the Listing/Releases/Publicise/etc tabs, or may it be reused elsewhere as well?

Seems like we would need to QA more than just the Publicise page to make sure there are no regressions.

bartaz:
Looks like Collaboration is broken, as it was using application layout for the "Add new collaborator" panel: https://charmhub-io-2082.demos.haus/landscape-keystone/collaboration

static/js/src/publisher-admin/routes/root.tsx
```diff
@@ -19,7 +20,25 @@ function Root() {
     }
   }, [packageData]);
 
-  return (
+  const isCollaborationTab = location.pathname.endsWith("/collaboration");
```
bartaz, line 23: That's very specific and quite fragile workaround.

A bit safer approach would be to move the application layout divs to the actual view of the collaboration tab which I guess is this one [Collaboration.tsx](https://github.com/canonical/charmhub.io/blob/main/static/js/src/publisher-admin/pages/Collaboration/Collaboration.tsx#L91) and wrapping contents there in the application layout stuff.
This will not be exactly the same as before (it will be a bit more nested), but from a quick check it seems to work and isolates the application layout a bit better (to the view that actually uses it).

static/js/src/publisher-admin/pages/Collaboration/Collaboration.tsx
```diff
@@ -89,7 +89,7 @@ function Collaboration() {
 
   return (
     <>
-      <div className="u-fixed-width">
+      <div className="l-application">
         {showRevokeSuccess && (
           <Notification
```
bartaz, line 94: I'm not sure how that would work for notifications (being under `l-application`, but not in any panel).
I believe it would be safer to keep them inside main panel, with rest of the content, or above `l-application` (still inside it's own `u-fixed-width`.

I'm kinda guessing here, not sure how to QA it. Do we have any testing charm to try to add collaborators and see if notification pops up?

abbiesims: @bartaz ah yep, it's off, I just tried it with a charm I'm a collaborator on. Let me fix it!
Screenshot from 2025-02-14 12-47-09

abbiesims: @bartaz invited you to the charm so you can QA, but it now looks like this:
Screenshot from 2025-02-14 12-58-36

PR #2088, opened 2025-02-19
add safe redirects for docs versions

abbiesims:
## Done
Adds safe redirects to avoid XSS.

## How to QA
Unfortunately, none of our docs have versions so the easiest way to QA is:
- Check out this branch
- Add the following dummy component to the `document.html` and `_side-nav.html` files:
```
<div id="test-urlencode">
  <h3>Redirect Test</h3>
  <label for="version-select">Test safe redirects:</label>
  <select id="version-select">
  <option value="">-- Select Test URL --</option>
  <option value="{{ '/docs/test?name=John Doe&role=admin' }}">Encoded Test 1</option>
  <option value="{{ '/docs/special characters & symbols' }}">Encoded Test 2</option>
  <option value="{{ 'javascript:alert(\'XSS\')' }}">Should Not Execute</option>
</select>
</div>
```
- Run the local host
- Then go to a topic (http://localhost:8045/topics/charmed-tempo-ha) and also a docs page (http://localhost:8045/kafka-k8s) and test the dropdowns
  - The first two links will return 404 but check the urls are correctly formatted
  - The final link should not redirect and there should be a warning in the console

## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): no change in current behaviour

## Issue / Card
Fixes https://github.com/canonical/charmhub.io/security/code-scanning/8 and https://github.com/canonical/charmhub.io/security/code-scanning/7

### DIFF
--- a/static/js/src/base/redirect-docs-url.ts
+++ b/static/js/src/base/redirect-docs-url.ts
@@ -1,0 +1,26 @@
+document.addEventListener("DOMContentLoaded", () => {
+  const selectElement = document.getElementById(
+    "version-select"
+  ) as HTMLSelectElement | null;
+
+  if (!selectElement) return;
+
+  selectElement.addEventListener("change", () => {
+    const url = selectElement.value.trim();
+
+    if (!url) return;
+
+    try {
+      const parsedUrl = new URL(url, window.location.origin);
+
+      // Allow only URLs starting with "http", "https", or "/"
+      if (/^(https?:\/\/|\/)/.test(parsedUrl.href)) {
+        window.location.href = parsedUrl.href;
+      } else {
+        console.error("Blocked unsafe redirect:", parsedUrl.href);
+      }
+    } catch (error) {
+      console.error("Failed to parse URL:", url, error);
+    }
+  });
+});
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "babel-loader": "9.2.1",
     "css-loader": "7.1.2",
     "date-fns": "4.1.0",
+    "dompurify": "^3.2.4",
     "expose-loader": "5.0.0",
     "identity-obj-proxy": "3.0.0",
     "mermaid": "11.4.1",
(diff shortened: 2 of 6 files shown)


static/js/src/base/redirect-docs-url.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (!selectElement) return;
+
+  selectElement.addEventListener("change", () => {
+    const url = selectElement.value.trim();
+
+    if (!url) return;
+
+    // Allow only URLs starting with "http", "https", or "/"
+    if (/^(https?:\/\/|\/)/.test(url)) {
+      const sanitizedUrl = DOMPurify.sanitize(url, {
```
M7mdisk, line 17: FYI I think copilot is being silly here, it makes no sense to use DOMPurify becaues the url is not a DOM element, neither is it being attached to our DOM, unless I'm missing something

static/js/src/base/redirect-docs-url.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (!selectElement) return;
+
+  selectElement.addEventListener("change", () => {
+    const url = selectElement.value.trim();
+
+    if (!url) return;
+
+    // Allow only URLs starting with "http", "https", or "/"
+    if (/^(https?:\/\/|\/)/.test(url)) {
+      window.location.href = url;
```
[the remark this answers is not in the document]
M7mdisk: referring to this ^

static/js/src/base/redirect-docs-url.ts
M7mdisk: I'd remove it, I've also noticed it hallucinating lots of security (non)-issues recently. Its good to take a look at what copilot suggests in case we miss something but this looks like a false positive

codeEmpress1:
The URL are not well formatted `http://localhost:8045/docs/test%3Fname%3DJohn%20Doe%26role%3Dadmin`

codeEmpress1:
LGTM!

codeEmpress1 (approved):

PR #2094, opened 2025-02-21
safely render svg

abbiesims:
## Done
Safely renders SVG by sanitising it to prevent XSS.

## How to QA
- I couldn't find any charm docs that render a mermaid diagram so the easiest way to QA is to check out this branch and add this test component to the bottom of the `mermaid.js` file:
```
// TEST COMPONENT TO CALL renderSVG
document.addEventListener("DOMContentLoaded", async () => {
  // Create fake Mermaid block
  const testBlock = document.createElement("div");
  testBlock.innerHTML = `<code>flowchart TD; A-->B</code>`;

  // Wrap in a parent div (to mimic actual DOM structure)
  const wrapper = document.createElement("div");
  wrapper.appendChild(testBlock);
  document.body.appendChild(wrapper);

  // Ensure the block is in the DOM before rendering
  await new Promise((resolve) => setTimeout(resolve, 100));

  // Call `renderSVG` to make sure the mermaid diagram renders in the UI
  await renderSVG(testBlock.firstChild, 0);
});
```
- Run localhost
- Go to a charm's docs page (e.g. localhost:8045/jenkins-agent/docs/reference-configurations)
- Check that the mermaid svg renders at the bottom
- Choose `markup` in the dropdown and ensure the code snippet renders

## Testing
- [ ] This PR has tests

## Issue / Card
Fixes https://github.com/canonical/charmhub.io/security/code-scanning/3

### DIFF
--- a/static/js/src/public/docs/mermaid.js
+++ b/static/js/src/public/docs/mermaid.js
@@ -1,4 +1,5 @@
 import mermaid from "mermaid";
+import DOMPurify from "dompurify";
 
 // We don't get the "mermaid" code block id through
 // so be really hacky about it
@@ -44,7 +45,12 @@ async function renderSVG(block, index) {
   const renderedDiv = document.createElement("div");
   renderedDiv.className = "rendered";
   renderedDiv.style.marginTop = "1rem";
-  renderedDiv.innerHTML = svg;
+
+  const sanitizedSVG = DOMPurify.sanitize(svg, {
+    ADD_TAGS: ["svg", "g", "path", "rect", "circle", "text", "foreignObject"],
+    ADD_ATTR: ["viewBox", "xmlns", "width", "height", "style", "class", "id"],
+  });
+  renderedDiv.innerHTML = sanitizedSVG;
 
   const markupDiv = document.createElement("div");
   markupDiv.className = "markup";


bartaz:
Not sure if it's related to PR changes, or to the test method (manually creating and attaching fake chart), but the markup seems to include HTML and escaped symbols, while I guess it should just show `flowchart TD; A-->B`?

image

But I don't know how it is expected to work on real example.

static/js/src/public/docs/mermaid.js
```diff
@@ -44,7 +45,12 @@ async function renderSVG(block, index) {
   const renderedDiv = document.createElement("div");
   renderedDiv.className = "rendered";
   renderedDiv.style.marginTop = "1rem";
-  renderedDiv.innerHTML = svg;
+
+  const sanitizedSVG = DOMPurify.sanitize(svg, {
+    ADD_TAGS: ["svg", "g", "path", "rect", "circle", "text", "foreignObject"],
```
bartaz, line 50: How do we know if this list covers everything mermaid can render?

static/js/src/public/docs/mermaid.js
```diff
@@ -44,7 +45,12 @@ async function renderSVG(block, index) {
   const renderedDiv = document.createElement("div");
   renderedDiv.className = "rendered";
   renderedDiv.style.marginTop = "1rem";
-  renderedDiv.innerHTML = svg;
+
+  const sanitizedSVG = DOMPurify.sanitize(svg, {
```
bartaz, line 49: In a way I wonder if we actually need this sanitation here.
The value of `svg` comes from mermaid call. It's of course safer to sanitize it, but we can assume it trust it. The output of it can't really be controlled by attacker.

The problematic part flagged by security check was related not to SVG I believe, but to `block.parentNode.outerHTML` being rendered as "Markup" part. And we still do it (in line 57), just pass it to `textContent` instead of `innerHTML` which causes the escaping issues I mentioned elsewhere.

So my understanding is that we could leave SVG without sanitation actually, but we need to address the rendering of Markup part somehow.

abbiesims:
> Not sure if it's related to PR changes, or to the test method (manually creating and attaching fake chart), but the markup seems to include HTML and escaped symbols, while I guess it should just show `flowchart TD; A-->B`?
> 
> image
> But I don't know how it is expected to work on real example.

Nice catch, will do some investigation to see if this is from the test method or the actual change

static/js/src/public/docs/mermaid.js
abbiesims: I realised I attached the wrong issue - it is actually [this alert](https://github.com/canonical/charmhub.io/security/code-scanning/35). I also presumed we could trust the output of a mermaid call but when I fixed the [original issue](https://github.com/canonical/charmhub.io/security/code-scanning/3), it then flagged this one

static/js/src/public/docs/mermaid.js
abbiesims: There will be a lot more types not included here, but I'm also thinking we could even reduce this to `svg` for now as that seems to be all we render throughout the repo. Again, also hard to test as I couldn't find anywhere that this exact component was used. What do you think?

bartaz: Looking at what mermaid can render in the charts it's a lot of possibilities.
We would likely need to allow **all** valid SVG elements.

I don't know how seriously we need to consider this particular warning, because the SVG contents itself are not user-controlled, they are out of mermaid library (which I guess we may assume we are trusting)?

Another approach could be to at least disallow some potentially malicious elements like `script` or `onclick` attribute, etc, to avoid any potential script injection.

Because right now we are simply risking that some charts will not render properly in unexpected ways.
I don't know what is the expected use case for those charts and do we want (and need) to limit them in any way.

bartaz: I found this: https://mermaid.js.org/config/usage.html#securitylevel

It seems by default it's "strict", so I beliave it should be OK for us to trust mermaid sanitisation.
Not sure how we can make the alert go away in such case though.

abbiesims: Okay nice, thanks! Will close this PR, and we can also dismiss the alert in the security tab.

PR #2095, opened 2025-02-21
encode redirect after accepting invite

abbiesims:
## Testing
- [ ] This PR has tests

## Issue / Card
Fixes #

### DIFF
--- a/templates/publisher/accept-invite.html
+++ b/templates/publisher/accept-invite.html
@@ -102,7 +102,7 @@
       inviteSuccessPanel.classList.remove("u-hide");
 
       setTimeout(() => {
-        window.location = `/${package}/listing`;
+        window.location = `/${encodeURIComponent(package)}/listing`;
       }, 3000);
     }).catch((error) => {
       closeAllNotifcations();


templates/publisher/accept-invite.html
```diff
@@ -102,7 +102,7 @@ <h5 class="p-notification__title">Error</h5>
       inviteSuccessPanel.classList.remove("u-hide");
 
       setTimeout(() => {
-        window.location = `/${package}/listing`;
+        window.location = `/${encodeURIComponent(package)}/listing`;
```
bartaz, line 105: This still seems to be flagged by code scanning.

We obviously can't just keep the list of all allowed packages. 
I wonder if the security check incorrectly assumes that this value is provided by the user? Or is the `package` something we control?

bartaz: So, we are reading `package` out of search params:

https://github.com/canonical/charmhub.io/blob/3d01f508085257bfce4e95767e59f48706cd7952/templates/publisher/accept-invite.html#L64

This means it is user-provided value.

I'm still not really sure how would that redirect be unsafe if we redirect within current host because URL starts with `/` - unless there is some obscure way to get around that?
Still, this only triggers after successful `/accept-invite` API call, which would fail if the `package` is invalid or malcious.

So my gut feeling is that our code is secure enough ATM.

I'm not sure if there is a way to get the `package` name from trusty source, not from URL params?
Does `accept-invite` API successful response include package name, so that we could take that instead?

abbiesims: Successful response is empty from the API, so unfortunately not much we can do with that. If you think it's already secure enough, we can dismiss the alert. Or we could still encode it for slightly better security, and then dismiss the resulting alert?

bartaz: Potential solution of changing the API is probably not worth the effort, as looking at the code and the flow of how it works it's highly unlikely this can actually be utilised in any malicious way.

I think we are fine merging this as is (with added value of encoding URL) and dismissing the warning. We may get back to API change as a solution if needed later.

bartaz (approved):
Approving as is.

This still triggers warning, but we believe it's secure enough in current form. May revisit if needed.

PR #2119, opened 2025-04-17
fix: disable talisker for charmed deployment

M7mdisk:
## Done
- disable talisker logging in charmed deployment (it is cluttering the logs with warnings about it being disabled)

## How to QA
- Trust :)
- As a hack: run locally with `dotrun exec` followed by `FLASK_SECRET_KEY=test flask run`
- See that the `track API called when there is no active context, data will not be stored` line is not showing


steverydz (approved):
LGTM 👍

PR #2130, opened 2025-05-08
feat: add tracing to store_api calls

M7mdisk:
## Done
- Make all store_api instances (traced) singletons.
- change exclude pattern to match any static files
## How to QA
- spin up o11y locally as per `observability/README.md`
- make a bunch of requests on charmhub (click around some charms, login)
- should see requests made to store_api
- you SHOULD NOT see any requests to static js, .css, images etc
## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): tracing

## Issue / Card
Fixes WD-21964

### DIFF
--- a/tests/publisher/test_publisher_views.py
+++ b/tests/publisher/test_publisher_views.py
@@ -45,10 +45,7 @@ class TestPublisherViews(unittest.TestCase):
         self.assertIn(b"test-email", res.data)
         self.assertIn(b"test-username", res.data)
 
-    @patch(
-        "canonicalwebteam.store_api.publishergw."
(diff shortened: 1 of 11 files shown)


alvaromateo (approved):
LGTM!
I have a suggestion for the `decorate_all_methods` method.
Given that maybe not all the attributes from the class will be methods to which we should add the tracing functionality, we could add the __dir__() method to PublisherGW and DeviceGW to return only the methods that we need to trace.
See https://docs.python.org/3/library/functions.html#dir.

abbiesims (approved):
LGTM! :+1:

PR #2137, opened 2025-06-02
CD: reusable deployment flow

M7mdisk:
## Done
- New unified deployment flow, deploys main on prod and staging to staging.
- This workflow is made to be reusable so that it can be shared across projects, all configuration options are environment variables set in the repo settings.
## How to QA
- [successful run](https://github.com/canonical/charmhub.io/actions/runs/15387105048) 
## Testing
- [ ] This PR has tests
- [x] No testing required (explain why): CD change


.github/workflows/deploy.yaml
```diff
@@ -1,30 +1,161 @@
...
+          name: ${{ env.CHARM_NAME }}-rock
+
+      - name: Set image URL
+        id: set_image_url
+        run: echo "image_url=ghcr.io/${GITHUB_REPOSITORY}:$(date +%s)-${GITHUB_SHA:0:7}" >> $GITHUB_OUTPUT
 
-  deploy-prod:
-    if: ${{ ( github.ref == 'refs/heads/main' && github.event.workflow_run.conclusion == 'success') || github.event_name == 'workflow_dispatch' }}
+      - name: Push to GHCR
+        run: skopeo --insecure-policy copy oci-archive:$(ls *.rock) docker://${{ steps.set_image_url.outputs.image_url }} --dest-creds "canonical:${{ secrets.GITHUB_TOKEN }}"
```
goulinkh, line 96: Nit: we make the part `canonical` from the credentials dynamic with something like `github.repository_owner`.

To be tested because I'm not 100% sure if it returns `canonical` in the case where the CI runs on the original repo and return the forked username in case of running action as a fork.

.github/workflows/deploy.yaml
```diff
@@ -1,30 +1,161 @@
...
+
+      - name: Pull image into registry
+        run: docker pull ${{ steps.set_image_url.outputs.image_url }}
+
+      - name: Upload flask app OCI image
```
goulinkh, line 101: This could be named in a more generic way not limit this job to Flask.

.github/workflows/deploy.yaml
```diff
@@ -1,30 +1,161 @@
...
+        env:
+          CHARMCRAFT_AUTH: ${{ secrets.CHARMHUB_TOKEN }}
+        run: |
+          sudo snap install charmcraft --classic
+          charmcraft upload-resource ${{ env.CHARM_NAME }} flask-app-image --image=$(docker images --format "{{.ID}}" | head -n 1) --verbosity=trace
```
goulinkh, line 106: This approach limits the CI job to public websites as Charmhub has not support for private Charms.

An alternative approach would be to reference the `.charm` file during the `juju refresh` command which will allow for private repos to use this CI job.

Example: https://github.com/canonical/specs-v2.canonical.com/blob/main/.github/workflows/deploy.yaml#L227

.github/workflows/deploy.yaml
M7mdisk: But do we want this to run on forks even? it should only run on pushes to the main repo anyway

.github/workflows/deploy.yaml
goulinkh: I see, I though all 12factor projects have the same resource name for the OCI image but it doesn't seem to be the case.
For the GH Action app we should probably provide per 12factor flavor action app and find a way to avoid writing duplicated YAML. Something like base template with inputs could work..

PR #2148, opened 2025-07-01
Collaborator Emails

M7mdisk:
## Done
- Send emails on collaborator invite
## How to QA
- go to https://staging.charmhub.io/charms
- pick a charm
- send an invite to a collaborator (use a real email)
- recieve email with invite link
## Testing
- [x] This PR has tests

### DIFF
--- a/static/js/src/publisher-admin/pages/Collaboration/InviteCollaborator.tsx
+++ b/static/js/src/publisher-admin/pages/Collaboration/InviteCollaborator.tsx
@@ -16,10 +16,12 @@ import {
   activeInviteEmailState,
   invitesListState,
   inviteLinkState,
+  publisherState,
 } from "../../state/atoms";
 
 import { generateInviteToken } from "../../utils/generateInviteToken";
 import { Invite } from "../../types";
+import { filteredCollaboratorsListState } from "../../state/selectors";
 
 type Props = {
   setShowSidePanel: (showSidePanel: boolean) => void;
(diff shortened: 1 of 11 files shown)


abbiesims:
nit: could we maybe add a success message when the user clicks "Send invite", after the loading state? feel free to do this in a separate PR if you think it's out of scope!

static/js/src/publisher-admin/pages/Collaboration/InviteCollaborator.tsx
```diff
@@ -36,28 +38,28 @@ function InviteCollaborator({ setShowSidePanel }: Props): JSX.Element {
...
 
-    const existingInvites = invitesList.filter((invite: Invite) => {
-      return invite.email === email && isPending(invite);
-    });
-
-    if (!existingInvites.length) {
+  const isValidEmail = (email: string | undefined) => {
+    if (!email) {
       return true;
     }
```
codeEmpress1, line 60: This part of the function would return true if the email is undefined or an empty string. Is this what is intended? I expect that undefined or empty string should return false that invalid email.

webapp/observability/metrics.py
```diff
@@ -0,0 +1,67 @@
+from typing import Dict
```
abbiesims, line 1: I think this file can go into the separate metrics PR? (sorry!)

abbiesims (approved):
All good apart from the one `metrics` file that can be moved! Nice work :rocket:

PR #2179, opened 2025-09-03
Wd 23472/add solutions tab

abbiesims:
## Done
- Adds Solutions tab to publisher pages if the logged in user has access to any published solutions

## How to QA
- Check out the solutions-service repo (`gh repo clone canonical/charmhub-solutions-service`)
- Run this using `docker compose up --build`
- Check this branch out locally and run `dotrun`
- Check that you don't see a solutions tab when you go to /charms
- Try and go to /solutions and make sure you are redirected to /charms
- Then, in charmhub-solutions-service/seed.py, change the publisher of a solution to `webteam-charmers`
  - e.g. change the publisher of the Observability solution to:
  ```
    "publisher": {
        "display_name": "Canonical Webteam Charmers",
        "id": "PsY0bciQRGBzO2dJoPlHjoE89LY6jGlD",
        "username": "webteam-charmers",
     },
  ```
- Stop the solutions service docker container and run `docker compose down -v` to clean the volumes
- Re-run it using `docker compose up --build`
- Refresh local charmhub and make sure you see the Solutions tab
- Go to the Solutions tab and make sure you see the COS solution

## Issue / Card
Fixes [WD-23472](https://warthogs.atlassian.net/browse/WD-23472)

## Screenshots

[WD-23472]: https://warthogs.atlassian.net/browse/WD-23472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webapp/solutions/logic.py
```diff
@@ -0,0 +1,60 @@
...
+
+    return []
+
+
+def publisher_has_solutions(username):
+    """
+    Checks if the user logged in has access to any solutions
+    - adds simple caching because it's quite slow right now
+    """
+    from flask import session as flask_session
```
M7mdisk, line 49: nit: why is the import here and not the start of the file?

webapp/solutions/logic.py
```diff
@@ -0,0 +1,60 @@
...
+        if resp.status_code == 200:
+            return resp.json()
+    except Exception:
+        pass
+    return None
+
+
+def get_publisher_solutions(username):
+    try:
+        token = login(username)
```
M7mdisk, line 26: Definitely should store the token in the session btw! Login makes a request to LP to get the teams, which alone takes ~2s. if the token expires, you can call login again and update the token.

webapp/solutions/logic.py
```diff
@@ -0,0 +1,60 @@
...
+        headers = {"Authorization": f"Bearer {token}"}
```
M7mdisk, line 27: nit: I assume we'll add more endpoints that require auth, maybe adding a helper that (1) logs in if not logged in (2) auto updates token if expired (3) adds auth header, would make our lives easier

webapp/publisher/views.py
```diff
@@ -125,6 +137,29 @@ def list_page():
     return render_template("publisher/list.html", **context)
 
 
+@trace_function
```
M7mdisk, line 140: just for future reference: no need to add `@trace_function` to routes, the flask instrumentation does it automatically

templates/publisher/solutions.html
```diff
@@ -0,0 +1,107 @@
...
+              <div class="p-heading-icon--x-small">
+                <span class="p-heading-icon__header">
+                  <p class="u-no-margin--bottom u-no-padding--top">
+                    <a href="/solutions/{{ solution.name }}">{{ solution.title }}</a>
+                  </p>
+                </span>
+              </div>
+            </td>
+            <td role="gridcell" aria-label="status">
+                {{ solution.status }}
```
M7mdisk, line 78: super-nit: (P)ublished

Image

webapp/solutions/logic.py
abbiesims: moved to top :+1:

webapp/solutions/logic.py
abbiesims: ahhh yes yes

M7mdisk (approved):
LGTM!
but I wouldn't merge until you get a +1 from someone else since we all had the same issue initially

edisile (approved):
LGTM, the only QA step I had to do differently was clearing my session cookie and logging in again before checking that the solution tab has appeared

PR #2180, opened 2025-09-04
add docker compose for redis

codeEmpress1:
## Done
- Add docker compose for running Redis locally
## Testing
- [ ] This PR has tests
- [x] No testing required (explain why):


alvaromateo (approved):
LGTM!
As a suggestion for the future: we could add to docker compose another docker container just with a slim ubuntu image with redis-cli installed, so that we can run `docker container exec -it container bash` and run redis-cli from there. That way we don't need to deal with different redis-cli versions for different OSs and we would all use the same.

codeEmpress1:
> LGTM! As a suggestion for the future: we could add to docker compose another docker container just with a slim ubuntu image with redis-cli installed, so that we can run `docker container exec -it container bash` and run redis-cli from there. That way we don't need to deal with different redis-cli versions for different OSs and we would all use the same.

Great suggestion Alvaro, I would look into it.

PR #2198, opened 2025-09-15
Wd 23478/add register solution

abbiesims:
## Done
- Adds logic and view to be able to register a new solution

## How to QA
- Check out this branch
  - Run `dotrun`
- Check out [this branch on the solutions service](https://github.com/canonical/charmhub-solutions-service/pull/17)
  - Run `docker compose up --build`
- Go to localhost:8045/solutions - you should see one published solution
- Click on "Register a new solution"
- Go through the process of registering a new solution
  - Try to make the form fail (invalid solution name, leave some input fields empty to make sure validation is working)
  - Click register when you have valid inputs
- You should see the new solution at `/solutions` with the status "Pending name review"
- If you go to `localhost:5000`, you should see the solution in the "pending name review" table
  - If you click "Approve" and go back to `localhost:8045/solutions`, the new solution should be in "Draft" status
- Preview button should work to show an empty solution page (need to remove some logic from this preview page because currently, links are pointing to nothing - will do this in a separate PR)

## Testing
- [x] This PR has tests

## Issue / Card
Fixes [WD-23478](https://warthogs.atlassian.net/browse/WD-23478)

[WD-23478]: https://warthogs.atlassian.net/browse/WD-23478?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/publisher/views.py
+++ b/webapp/publisher/views.py
@@ -641,14 +641,17 @@ def submit_register_solution():
     username = session["account"]["username"]
 
     form_data = {
-        "name": request.form.get("name", "").strip(),
-        "title": request.form.get("title", "").strip(),
-        "summary": request.form.get("summary", "").strip(),
-        "publisher": request.form.get("publisher", "").strip(),
-        "platform": request.form.get("platform", "kubernetes"),
-        "creator_email": request.form.get("creator_email", "").strip(),
-        "mattermost_handle": request.form.get("mattermost_handle", "").strip(),
-        "matrix_handle": request.form.get("matrix_handle", "").strip(),
+        field: request.form.get(field, "").strip()
+        for field in [
+            "name",
+            "title",
+            "summary",
+            "publisher",
+            "platform",
+            "creator_email",
+            "mattermost_handle",
+            "matrix_handle",
+        ]
     }
 
     user_teams = get_user_teams_for_solutions(username)
--- a/webapp/publisher/views.py
+++ b/webapp/publisher/views.py
@@ -19,7 +19,20 @@ from webapp.utils.emailer import get_emailer
 from webapp.solutions.logic import (
     get_publisher_solutions,
     publisher_has_solutions,
+    register_solution,
+    get_user_teams_for_solutions,
 )
+
+
+def _check_solutions_access(username):
+    """
+    Check if user has solutions access
+    Redirect to /charms if not
+    """
+    has_solutions = publisher_has_solutions(username)
+    if not has_solutions:
+        return redirect("/charms")
+    return True
 
 
 publisher = Blueprint(
@@ -143,16 +156,15 @@ def list_page():
 def solutions_page():
     username = session["account"]["username"]
 
-    # Check if user has access to solutions - redirect to /charms if not
-    has_solutions = publisher_has_solutions(username)
-    if not has_solutions:
-        return redirect("/charms")
+    access_check = _check_solutions_access(username)
+    if access_check is not True:
+        return access_check
 
     # Get solutions data according to launchpad group
     solutions = get_publisher_solutions(username)
 
     context = {
-        "has_solutions": has_solutions,
+        "has_solutions": True,
         "solutions": solutions,
         "page_type": "solution",
     }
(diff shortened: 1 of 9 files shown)


webapp/publisher/views.py
```diff
@@ -19,9 +19,22 @@
...
 
 
+def _check_solutions_access(username):
+    """
+    Check if user has solutions access
+    Redirect to /charms if not
+    """
+    has_solutions = publisher_has_solutions(username)
+    if not has_solutions:
+        return redirect("/charms")
```
codeEmpress1, line 34: Is there a way to let users know the reason why they're being redirected to charms? I think this might be a bit confusing for users, they might not know the reason they are being redirect to charms is because they do not have solutions.

webapp/publisher/views.py
```diff
@@ -143,16 +156,15 @@ def list_page():
 def solutions_page():
     username = session["account"]["username"]
 
-    # Check if user has access to solutions - redirect to /charms if not
-    has_solutions = publisher_has_solutions(username)
-    if not has_solutions:
-        return redirect("/charms")
+    access_check = _check_solutions_access(username)
+    if access_check is not True:
```
bartaz, line 160: I don't have an alternative proposal from the top of my head, but this seems like a bit of a weird pattern to have a function that returns either `True` or a redirect.

I understand that it tries to keep the redirect logic in single place, so it's not duplicated, but at the same time, the `if access_check is not True: return access_check` is duplicated in all places `_check_solutions_access` is used.

I feel that a conditional redirect is not something uncommon, so maybe there are some best practices for that in flask.

M7mdisk: I agree with bartek that this is kind of weird. I think  a nice way of doing this is would be through decorators `@has_solution_access` or `@solution_access_required`

Then something like

```python

def solutions_access_required(f):
    @wraps(f)
    def decorated_function(*args, **kwargs):
        username = session["account"]["username"]
        if not publisher_has_solutions(username):
            return redirect("/charms")
        return f(*args, **kwargs)
    return decorated_function
```

webapp/publisher/views.py
```diff
@@ -143,16 +156,15 @@ def list_page():
...
+        return access_check
 
     # Get solutions data according to launchpad group
     solutions = get_publisher_solutions(username)
 
     context = {
-        "has_solutions": has_solutions,
+        "has_solutions": True,
```
M7mdisk, line 167: A cool thing you can do here (I think)  is add stuff to the context based on the result of the decorator

```python
from flask import g
# ...right after the check

g.has_solutions = True
# ...
```

and then I think you can access that object in templates (again, not sure but try it) and you won't need to add has_solutions to the context everywhere

webapp/solutions/logic.py
```diff
@@ -81,3 +81,61 @@ def publisher_has_solutions(username):
...
     solutions = get_publisher_solutions(username)
     return bool(solutions and len(solutions) > 0)
+
+
+def register_solution(username, data):
+    try:
+        resp = make_authenticated_request(
+            "POST",
+            f"{SOLUTIONS_API_BASE}/publisher/solutions",
+            username,
```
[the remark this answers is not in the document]
abbiesims: Made a [ticket](https://warthogs.atlassian.net/browse/WD-27097) - will do this next week :+1:

webapp/publisher/views.py
abbiesims: Nice suggestion, thanks! Have done this for the functions that need the solutions decorator, but have kept it in the context for `list_page` because it doesn't use this decorator :+1:

edisile (approved):
LGTM!

webapp/publisher/views.py
```diff
@@ -19,7 +20,31 @@
...
+                "Launchpad team membership.",
+                "negative",
+            )
+            return redirect("/charms")
+
+        g.has_solutions = True
+        response = make_response(func(*args, **kwargs))
+        return response
+
+    return has_solutions_access
```
M7mdisk, line 47: lovely 👏

M7mdisk (approved):
Nice work!

PR #2240, opened 2025-11-13
WD-19723 Add sliding navigation

alvaromateo:
## Done

Add sliding navigation to charmhub.io.

## How to QA

- Go to [Demos](https://charmhub-io-2240.demos.haus/) and check that the navigation works the same way it does as in snapcraft.io.

## Testing
- [x] This PR has tests

## Issue / Card

Fixes [WD-19723](https://warthogs.atlassian.net/browse/WD-19723)

[WD-19723]: https://warthogs.atlassian.net/browse/WD-19723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/src/base/navigation/__tests__/navigationTemplate.ts
+++ b/static/js/src/base/navigation/__tests__/navigationTemplate.ts
@@ -1,228 +1,30 @@
-export const getNavigationTemplate = () => {
-  const topDiv = document.createElement("div");
-  // Keep the innerHTML in sync with the contents of templates/partial/_navigation.html
-  // so that the tests are meaningful
-  topDiv.innerHTML = `
-    <header id="navigation" class="p-navigation--sliding is-dark">
-      <div class="p-navigation__row--25-75">
-        <div class="p-navigation__banner">
-          <div class="p-navigation__tagged-logo">
-            <a class="p-navigation__link" href="https://juju.is">
-              <div class="p-navigation__logo-tag">
-                <img class="p-navigation__logo-icon" src="https://assets.ubuntu.com/v1/a603c7c9-Favicon - Juju.svg" alt="Canonical Juju" />
-              </div>
-              <span class="p-navigation__logo-title">Canonical Juju</span>
-            </a>
-          </div>
-          <ul class="p-navigation__items">
-            <li class="p-navigation__item">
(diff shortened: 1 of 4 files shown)


static/js/src/base/navigation/__tests__/navigationTemplate.ts
```diff
@@ -0,0 +1,231 @@
+export const getNavigationTemplate = () => {
+  const topDiv = document.createElement("div");
+  // Keep the innerHTML in sync with the contents of templates/partial/_navigation.html
```
bartaz: This seems extremely fragile. Especially for All Canonical global nav.

We do control our own navigation template, so technically we could keep it in sync with tests (if we remember to do so). But we don't control the output of global nav, as it's an external package.

So even if we have this test passing on the mocked data, it doesn't really guarantee that it will work on production with actual template or actual global nav.

Obviously its better to have a test than not having it at all, but II wonder if there are better ways to test it.

Integration tests seem better suited for such thing, as they would operate on actual website, but I don't know if we have a setup for those in charmhub at all.

alvaromateo: You're totally right @bartaz.

I've modified the navigationTemplate.ts file to pick the actual __navigation.html_ file and load that into the `document.body.innerHTML`; then call the `createNav()` function from global-nav module. This way we ensure we have always the same setup as in production and we can focus on testing the actual code that was added (adding the back buttons and the login method).

bartaz (approved):
The amount of duplication between snapcraft and charmhub worries me a bit. In the end it's hundreds of lines of code across multiple files.

Especially the TS scripts are quite complex, nuanced, and depend on HTML structure quite a lot, so it may be a hard thing to maintain.

Navigation does not change too often, so maybe it will not be too bad. Maybe this will be enough for us until navigation from new DS is ready.
But as well, with rocks store being worked on, I'd rather avoid duplicating the nav code once again there. Could be worth thinking of some upstreaming solution once we need that.

bartaz:
Demos don't seem to work (for final QA).

PR #2313, opened 2026-04-16
Wd 31018/refresh package details header

abbiesims:
## Done
- Refactors package details header to align with snapcraft.io
- Moves header logic to one directory to keep it organised
- Modifies visuals to align with snapcraft package header, according to [these designs](https://www.figma.com/design/P51yC6CoxDnGPzUDYDegih/Snaps-and-Charms---Sites?node-id=3-5&t=FYXzBzVJ9wnxNilF-0) ("After" section)

## How to QA
- Go to multiple charms in the demo (k8s and VM ones)
  - e.g. pgbouncer, kafka, mongodb, postgresql, postgresql-k8s, mysql-router-k8s, pgbouncer-k8s
- Make sure they match the [designs on figma](https://www.figma.com/design/P51yC6CoxDnGPzUDYDegih/Snaps-and-Charms---Sites?node-id=3-5&t=cltI9ucRj17iapaH-0)
  - background should all be grey like on snapcraft.io (previously, the header strip was grey and the rest of the page was white)
  - package title, channel map and tabs should all be aligned, with the logo on its own on the left
  - updates to platform version chips

## Testing

- [ ] This PR has tests
- [x] No testing required (explain why): only visual changes

## Issue / Card

Fixes [WD-31018](https://warthogs.atlassian.net/browse/WD-31018)

## UX Approval

- [ ] This PR does not require UX approval
- [x] This PR does require UX approval (add context):
  - see epics [here](https://warthogs.atlassian.net/browse/WD-29416) and [here](https://warthogs.atlassian.net/browse/WD-20984) & [design epic](https://warthogs.atlassian.net/browse/WD-29404)

[WD-31018]: https://warthogs.atlassian.net/browse/WD-31018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/details/_details-header.html
```diff
@@ -55,12 +55,16 @@
                 {{ platform }}
               {% endif %}
             </div>
-            <div>
+            <div class="series-tags u-no-margin--top">
```
steverydz: I don't think we need either class here as the `series-tags` class adds a top margin, and then we remove the margin with `u-no-margin--top`.

steverydz (approved):
LGTM 👍 

Just one small comment in the code but not a blocker

ilayda-cp (approved):

bartaz:
I'm not sure why (if we use same macros), but it seems to behave differently on smaller screens.

The title lands under logo, and it starts to look a bit messy:

image

image

Comparing to snapcraft, that preserves the icon next to title on smaller screens:

image

templates/details/package_header/_package_header.html
```diff
@@ -0,0 +1,70 @@
...
+                      fill=True,
+                      attrs={"class": "p-media-object__image"},) | safe
+              }}
+            {% endif %}
+          </div>
+        </div>
+      {% endif %}
+    </div>
+
+    <div class="grid-col-4">
```
bartaz: Looks like this part was modified in comparison to snapcraft (to add additional slot for platforms. But this doesn't provide mobile/tablet class names).

https://github.com/canonical/snapcraft.io/blob/9755ea144afa29642b4a898f465182b243bd0489/templates/store/package_header/_package_header.html#L31

abbiesims: yeah I had to make some modifications to allow for the platform version section that we don't have on snapcraft. will take a look and see if I can make this look nicer again

abbiesims:
@bartaz added back in grids for smaller screens, thanks for catching that. let me know what you think

bartaz:
It seems there is still quite a bit custom charmhub code on top of package header from snapcraft.

On mobile small screens it's quite visible, that logo image and title start to overlap, while they should be properly enclosed by their grid cells. Some (potentially unnecessary?) styles from charmhub are still being applied.
It would be great if we could try to remove any unnecessary divs, classes, etc, to make charmhub and snapcraft headers as close as possible, only keeping changes where necessary (platform area, install instructions, etc)

image

image

templates/details/package_header/_package_header.html
```diff
@@ -0,0 +1,70 @@
...
+  {%- set channel_map_content = caller("channel_map") -%}
+  {%- set has_channel_map_content = channel_map_content|trim|length > 0 -%}
+  {%- set platforms_content = caller("platforms") -%}
+  {%- set has_platforms_content = platforms_content|trim|length > 0 -%}
+  {%- set has_icon = package.type == "charm" -%}
+
+  <div class="grid-row">
+    <div class="grid-col-2 grid-col-medium-1 grid-col-small-1">
+      {% if has_icon %}
+        <div class="p-charm-header">
```
bartaz: These `p-charm-header` and `p-media-object--large` are not used in snapcraft. Do we need them here?
It would be great to stay as close as possible to snapcraft macros, so eventually we could technically extract them and only pass them different data.

templates/details/package_header/_package_header.html
```diff
@@ -0,0 +1,70 @@
...
+{%- macro package_header(package, developer_validation, image) -%}
+  {%- set channel_map_content = caller("channel_map") -%}
+  {%- set has_channel_map_content = channel_map_content|trim|length > 0 -%}
+  {%- set platforms_content = caller("platforms") -%}
+  {%- set has_platforms_content = platforms_content|trim|length > 0 -%}
+  {%- set has_icon = package.type == "charm" -%}
+
+  <div class="grid-row">
+    <div class="grid-col-2 grid-col-medium-1 grid-col-small-1">
+      {% if has_icon %}
```
bartaz: Any reason why this code is hardcoded here (as part of the `package_header` and not uncluded from a `package_icon` macro)?

I hoped we would be able to reuse the snapcraft macros "as is" without too much of changes, have there been any issues that caused it to rewrite quite large chunks of them? It seems to defeat the purpose a bit (at least on technical side).

bartaz (changes requested):
Would it be possible to get better alignment with snapcraft macros?

lukasSerelis:
The tooltip is unnecessary, as the link of learn to deploy on juju does the same thing.

Can we sneak in a copy button?
image

templates/details/package_header/_package_header.html
abbiesims: Yeah sorry - was so focused on changing the layout of things (because of extra fields in charmhub) that I strayed away from snapcraft. Have tried to align it back as much as possible :+1:

bartaz:
Bundles with no icon look a bit off on tablet and mobile because of empty space.

image
image

I guess the grid class names would need additional conditional values if there is an icon or not. That would make it 4 versions:
- with both icon and meta (platrofms)
- with icon and no meta
- no icon and meta
- no icon and no meta

templates/details/libraries/libraries_layout.html
```diff
@@ -11,8 +11,8 @@
           Toggle side navigation
         </a>
         <div class="p-side-navigation__overlay js-drawer-toggle" aria-controls="drawer"></div>
-        <nav class="p-side-navigation__drawer" aria-label="Example side navigation">
-          <div class="p-side-navigation__drawer-header">
+        <nav class="p-side-navigation__drawer u-no-padding--top u-no-padding--bottom p-strip--light" aria-label="Example side navigation">
```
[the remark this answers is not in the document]
bartaz: This indeed looks like an overlooked copy paste from Vanilla example.

templates/details/libraries/libraries_layout.html
```diff
@@ -11,8 +11,8 @@
           Toggle side navigation
         </a>
         <div class="p-side-navigation__overlay js-drawer-toggle" aria-controls="drawer"></div>
-        <nav class="p-side-navigation__drawer" aria-label="Example side navigation">
-          <div class="p-side-navigation__drawer-header">
+        <nav class="p-side-navigation__drawer u-no-padding--top u-no-padding--bottom p-strip--light" aria-label="Libraries side navigation">
```
bartaz: Why were the strip classes needed? The light strip was deprecated I think…

bartaz:
There is something off with side navigation (on smaller screens).

On prod it works correctly.

image

But on demos it's just a small overlay, not full screen:

image


On top of that, the side nav in "Libraries" tab doesn't work at all, and leaves a lot of space and some artifacts on the left side:

image

(on prod it also doesn't work, but it does not have excessive white space.

templates/details/libraries/libraries_layout.html
abbiesims: it is white without the `p-strip--light` class
image

bartaz: It seems that side nav has somehow different implementation on libraries and home page. And now they are both kinda broken (at least on smaller screen). I'll try to have a look later…

Wait, we don't have "is-paper" class on body?
Ah, that may be the issue. The light strip is used around whole main content. It should not -it has the wrong grey background.
We need `is-light` on body.
If we don't want to introduce it on all pages (to avoid regressions) we could limit it to details page in the template.

templates/details/_details-header.html
```diff
@@ -1,16 +1,16 @@
-{% from "details/_package_header.html" import package_header %}
+{% from "details/package_header/_package_header.html" import package_header with context %}
 
 <div class="p-strip--light is-shallow">
```
bartaz: This `--light` variant should not be used.

templates/details/libraries/libraries_layout.html
```diff
@@ -11,7 +11,7 @@
           Toggle side navigation
         </a>
         <div class="p-side-navigation__overlay js-drawer-toggle" aria-controls="drawer"></div>
-        <nav class="p-side-navigation__drawer" aria-label="Example side navigation">
+        <nav class="p-side-navigation__drawer u-no-padding--top u-no-padding--bottom" aria-label="Libraries side navigation">
```
bartaz: I believe the no-padding classes were added to accompany the strip light that we removed, so maybe they are not needed as well here?

bartaz (approved):
LGTM

(small comment about potentially unnecessary class names).

PR #2327, opened 2026-05-19
inline all JS

M7mdisk:
## Testing

- [ ] This PR has tests

## Issue / Card

Fixes #

## UX Approval

- [ ] This PR does not require UX approval
- [ ] This PR does require UX approval (add context):

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -9,7 +9,7 @@ from canonicalwebteam.candid import CandidClient
 from canonicalwebteam.flask_base.app import FlaskBase
 
 from webapp.store_api import publisher_gateway
-from webapp.extensions import csrf, vite
+from webapp.extensions import csrf, vite, vite_import
 from webapp.config import APP_NAME, VITE_CONFIG
 from webapp.handlers import set_handlers
 from webapp.login.views import login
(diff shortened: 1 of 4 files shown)


alvaromateo (commented):
I am wondering if we should make these changes inside _flask_vite_ directly, but given that we need to get this working soon I imagine a task could be created to upstream it - if everyone thinks like this (maybe @edisile opinion as creator of _flask_vite_ would be an important one to consider).

Regarding the code I agree with Copilot comments that CSP nonce is not used - and even if we set it up, given that we are not 100% SSR (due to content-cache), it is not the best approach for us.

I do not know how much of a security exposure is to put the JS inside `data:`. Copilot marked it as Medium and considering that we are in control of what we put I guess it is not critical. But I wonder if an attacker could make use of this `data:` to XSS us, given that it will be allowed by CSP.

I personally didn't find very clear the `rewrite_module_specifiers` function and Regex, but I guess it's something related to how Vite handles the imports (an explanation comment would be nice - but not critical).

alvaromateo (approved):
I tested and I don't see any JS in the network tab, so the code seems to be working.

I'm not sure we need the complexity of the import map and maybe that part can be removed. I gave a quick look at the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script/type/importmap) and it looks like it's just an alias map for imports. We are including the `<script type="module" ...` with all the JS code which I think it is the main reason why it works; and searching in the page I don't see any reference for the aliases in the import maps.

PR #2345, opened 2026-06-05
remove default-release search field

M7mdisk:
## Done
- as far as I can tell, the information from this field is not being used anywhere, so its removed.
## How to QA
- Find any place where it is used (or get your AI to do it) and get brownie points


steverydz (approved):
LGTM 👍

abbiesims:
I found a slight difference on `Integrations` tabs

On prod, we see the channel on integrations cards:
image

On demos, this is removed (although for some reason I don't ever remember seeing this in the past, and I'm not sure how important it is / useful it is?):
image

PR #2351, opened 2026-06-09
feat: integrations tab in details page

M7mdisk:
Adds integrations tab to the charm details page.
## Deviations:
- Mandatory is "Required"
# Q/A
- https://charmhub-io-2351.demos.haus/kafka/integrations 
- [Figma](https://www.figma.com/design/bY5HHkqzvYREdxJFhbl5kM/25.10-Charmhub-solution-focus---Store?node-id=1744-6566&m=dev)

### DIFF
--- a/static/js/src/public/details/integrations/components/App/App.tsx
+++ b/static/js/src/public/details/integrations/components/App/App.tsx
@@ -7,7 +7,6 @@ import {
   Col,
   Spinner,
   SearchAndFilter,
-  Chip,
 } from "@canonical/react-components";
 import { InterfaceItem } from "../InterfaceItem";
 import { filterChipsSelector, filterState } from "../../state";
--- a/static/js/src/public/details/integrations/components/App/App.tsx
+++ b/static/js/src/public/details/integrations/components/App/App.tsx
@@ -149,14 +149,6 @@ export const App = () => {
                     }}
                   >
                     {`${interfaceItem.key}`}
-                    {interfaceItem.required === true && (
-                      <Chip
-                        value="Required"
-                        appearance="negative"
-                        className="u-no-margin--bottom"
-                        style={{ marginLeft: "10px" }}
-                      />
-                    )}
                   </a>
                 </li>
               );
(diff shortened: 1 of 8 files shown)


static/js/src/public/details/integrations/components/App/App.tsx
```diff
@@ -197,40 +188,35 @@ export const App = () => {
...
-                <h2 className="p-heading--3 p-details-tab__content__body__title">
-                  {integrationCount} integration
-                  {integrationCount > 1 ? "s" : ""}
-                </h2>
-                <p className="p-heading--4 p-details-tab__content__body__link">
-                  <a href="https://juju.is/docs/juju/relation">
-                    Learn about integrations&nbsp;&gt;
-                  </a>
-                </p>
+                <Row className="u-flex p-divider p-details-tab__integrations-header-title">
```
steverydz: You shouldn't use `u-flex` on a `Row` component as it will interfere with the default grid styles. I think this should be a nested `div` in this case.

static/js/src/public/details/integrations/components/App/App.tsx
```diff
@@ -197,40 +188,35 @@ export const App = () => {
...
-                          prev !== searchData
-                            ? (searchData as IFilterChip[])
-                            : prev
-                        );
-                      }}
-                    />
-                  </div>
+                <div className="p-details-tab__integrations-search">
+                  <SearchAndFilter
+                    // @ts-expect-error: id mismatch (number instead of string) but doesn't matter in reality
```
steverydz: perhaps it's better to convert this number to a string rather than ignore the error?

static/js/src/public/details/integrations/components/InerfaceRow.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const packageLink = `/${charm.package.name}`;
+  const channel = getChannel(charm);
+  const supportsVm = charm.package.platforms.includes("vm");
+  const supportsKubernetes =
+    charm.package.platforms.includes("kubernetes") ||
+    charm.package.platforms.includes("k8s");
+
+  return (
+    <Row>
+      <Col size={3} className="u-flex">
```
steverydz: You shouldn't use `u-flex` on a `Col` component as it will interfere with the default grid styles. I think this should be a nested `div` in this case.

steverydz (changes requested):
Looks good but a couple of comments in the code

M7mdisk:
@steverydz thanks for the review! comments addressed

steverydz (approved):
LGTM 👍 

Thanks for the changes!

PR #2352, opened 2026-06-09
feat: interfaces table

M7mdisk:
Adds the interfaces index table and points to the new interfaces repo.
## Deviations:
- Version isn't a dropdown (instead the library/docs link has the top level info containing all versions)
- No charm filter (Adding this requires N+1 queries, and you can go to the /charm-name/integrations tab to get the same info)
- Mandatory is "Required" (Consistent with past label name)
# Q/A
- https://charmhub-io-2352.demos.haus/integrations
- [Figma](https://www.figma.com/design/bY5HHkqzvYREdxJFhbl5kM/25.10-Charmhub-solution-focus---Store?node-id=1744-5226&m=dev)


steverydz (approved):
LGTM 👍

PR #2353, opened 2026-06-09
chore: add interface links to homepage

M7mdisk:
Adds interface links to the homepage 

# QA
- Design matches [Figma](https://www.figma.com/design/bY5HHkqzvYREdxJFhbl5kM/25.10-Charmhub-solution-focus---Store?node-id=1749-39690&t=g8IRHDMmGodMUpNt-4)

PR #2356, opened 2026-06-10
Interfaces

M7mdisk:
## Done
Add all integrations work per [designs](https://www.figma.com/design/bY5HHkqzvYREdxJFhbl5kM/25.10-Charmhub-solution-focus---Store?node-id=1749-39690). All code has been reviewed in #2352 #2353 and #2351.
### Deviations
- Version isn't a dropdown (instead the library/docs link has the top level info containing all versions)
- No charm filter (Adding this requires N+1 queries, and you can go to the /charm-name/integrations tab to get the same info)
- Mandatory is "Required" (Consistent with past label name)
## How to QA
- https://charmhub-io-2356.demos.haus/
- https://charmhub-io-2356.demos.haus/integrations
- https://charmhub-io-2356.demos.haus/kafka/integrations

## Testing

- [x] This PR has tests

## Issue / Card

Fixes WD-36229

## UX Approval
- [X] This PR does require UX approval (add context):


migueldivo:
Hi @M7mdisk, this should be reviewed as well by @lukasSerelis, but some immediate things I noticed are the following:

In the Charmhub header/hero links, Lukas' design includes an icon for the second link. Would you be able to add this icon as it help communicate that this link takes the user to an external page.
image


The pagination in the interfaces table page should be aligned to the right side, currently its aligned to the left
image


Maybe this is a rendering problem from my end, but in the example integration tab for Kafka, the overview number, divider, and link seem to be overlapping 
image

james-garner-canonical:
Wow, this looks great, thanks for sharing @M7mdisk! I'm going to post a separate comment for each page to make replies easier.

james-garner-canonical:
> https://charmhub-io-2356.demos.haus/

I'm a bit hesitant about the current 'Developer Resources' link. They look very tasteful thanks to @lukasSerelis's designs, if perhaps a bit _too_ prominent (they push the charm content down a fair bit with the current spacing). But my main hesitation is that I don't think Charmlibs is really the right link here. If anything, this should be a link to some sort of entrypoint for writing charms, but this is something that there still isn't a satisfactory location to link to (docs folks have some work in flight currently).

I'd recommend dropping this section for now (perhaps placing the 'Interfaces' link in the footer, or accepting that the main navigation target will be from specific interface pages).

However, if we're committed to this design, then "Interfaces | Write a charm \<external link icon>" could work, linking to the Ops docs homepage for how (which has prominent pointers to both 12-factor charming tutorials and tutorials for charming directory with Ops).

james-garner-canonical:
> https://charmhub-io-2356.demos.haus/integrations

The updated table looks great, so much more informative than the old version. I have a few comments on the different columns.

- Rename the 'Interface name' column to 'Interface' (more obvious that there's a useful link there).
- An interface can have more than one category tag, so the 'Category' column should be named 'Categories' (or 'Tags'?).
- The latest interface version number isn't very useful to surface here, particularly with the new approach for interfaces being strict backwards compatibility -- how about we drop this column?
- The links column ends up being a little confusing for users, I think -- it's not obvious that 'Library' and 'Documentation' link to different places, or what the documentation is for (library or interface). I recommend we drop this column entirely (and update the specific interface pages to have these links at the top). However, if we're committed to having this column, I think it would be a lot clearer to only have the library link.

Also dropping pagination would be nice! The table is cheap to construct (generated from a single JSON file), and there isn't any important content below it. The Figma design doesn't show pagination either.

> Most of the content of these pages can be collaboratively discussed and changed. You can help us improve these pages by clicking the button below.

This text at the top of the page, and the contribute button below it, are outdated. I think they should just be dropped entirely -- the contribution flow will be via the Charmlibs content, which we'll link to individually.

james-garner-canonical:
> https://charmhub-io-2356.demos.haus/kafka/integrations

I like the new look. Based on discussion with our technical author, I wonder if we could streamline the (currently unchanged) auto-generated text on this page more generally. We found it a bit hard to parse as-is.

1. Change the RHS text from "The <endpoint name> endpoint <provides|requires> an integration over the <interface name> interface" to "This charm <provides|requires> the <interface name> interface."
2. Drop the large `<interface name> interface` text below the `<endpoint name> endpoint <required>` text.
3. Change "This means it can integrate with:" conditionally to either "It can integrate with:" or "No publicly listed charms <require|provide> this interface." (where the role is the opposite of the charm's, addressing a longstanding and still-present bug that the role is the same).
4. Drop the separete "No charms found ..." text below.

Here's a mock up using developer tools:

Screenshot From 2026-06-16 17-19-33

---

Separately, the text at the top currently doesn't quite match the formatting proposed in the Figma designs (no separation, separator is overlayed on text). But I wonder if we should really link to the integrations list page from here -- mightn't this be better placed on the interface pages themselves? We could focus this text on the charm user, perhaps something like: "This charm has 8 endpoints for integrating with other charms."

dwilding:
Sorry for the earlier ping on this PR - I posted a comment by accident. This one is a real comment!

I agree with the adjustments that James is proposing.

I've been thinking more about the "Developer resources" links during the day and I prefer that we remove it for now. As James mentioned, we're internally working through the question of where we point people who are looking to get into charm development. Once we have a better answer to that, I think we can advertise it on Charmhub.

Removing "Developer resources" would mean lower discoverability of the `/integrations` page. I'm OK with that. I feel it shouldn't need to be a starting point. I think this is a perfectly fine flow:

```
/mongodb/integrations  I'm exploring the integrations that a charm supports,
                       then click on one of the interface links.
                       For example, 'shards'
   ↓
/integrations/shards   Here I can get a summary of how the interface works.
                       I should also be able to navigate up to see all interfaces
   ↓
/integrations
```

I don't think this PR covers changes to the `/integrations/foo` pages, does it? At very minimum we'd want to make sure there's a link up to `/integrations`.

james-garner-canonical:
Thanks @dwilding. Good point about the `/integrations/foo` pages, which indeed are unchanged in this PR right now. Here's my sketch for what we could do there, as I do think it's an important part of the whole picture.

- Add a prominent backlink to the `integrations/` page.
- Don't embed developer documentation in a section below the 'Charms' section.
- Instead embed the longer `description` field from Charmlibs at the top of the page. I'd also repeat the 'summary' from the index table somewhere on the page close to the interface name.
- Follow the description with a few sentences:
  - "The charms that can be related over this interface are listed below." -- a sentence for charm users.
  - If the `lib` field is populated, add a sentence for charm authors: "To use this interface in your charm, use the <lib name> library." if the `lib_url` field is populated, make <lib name> a link.
    - If the (new) `lib_docs_url` field is present: "See the library docs.", with 'library docs' linking to `lib_docs_url`.
  - Finally if the `docs_url` field is present: "To learn more about how this interface works under the hood, read the interface docs.", with 'interface docs' linking to `docs_url`.

(Here's a link to the preview for this page in the current PR, unchanged compared to the live version at time of writing: https://charmhub-io-2356.demos.haus/integrations/cos_agent).

migueldivo:
Thanks for the all the feedback. From a UX perspective I've got some comments to address before we move forward with any changes


### Charmhub landing page 'Developer Resource' block
I agree that if the documentation that is proposed to be linked in hero/header section are not the correct places we want people to onboard into, we shouldn't be highlighting it so prominently. 
So here are my suggestions:
- I agree that including the 'Developer Resources' in the footer would be a good workaround for as it can still be discoverable on every charmhub page
- To ensure that the interfaces page is not an unreachable page, I suggest we add it in under the category filters. This makes sure that we're not making users traverse backwards to find these links and they're not prominently showcased at the top of the page and not hidden away in the footer. This also means that we don't need to add a link back to` /integrations` from` /integrations/foo` (as we also don't have an approved design pattern/component for this, we rely on the user going back a page on the browser)
image

- I like the suggestion of having the 'write a charm' link under 'Developer Resources' as it works nicely for onboarding people into creating their own charms. 
- We should set the links to a maximum of 4 so that we aren't overwhelming users with so many external links on charmhub 



 ### Interfaces index/table [page](https://charmhub-io-2356.demos.haus/integrations)
 No problems with renaming the table columns to 'Interface' and 'Categories' (I wouldn't change categories to tags because on charmhub we refer to groupings/tags as categories as per the current charm filtering).
 
 Its great that the interface name is linked to the individual interface page for people to discover all the libraries and details of that interface. Where I'm seeing an issue with the 'Links' column is that an individual interface can have multiple libraries linked to it as they are either providing or requiring that respective interface to multiple charms. E.g. `cos_agent`
on the [interface table](https://charmhub-io-2356.demos.haus/integrations), the 'Library' link for `cos_agent` links to the library tab on the [Grafana agent charm](https://charmhub.io/grafana-agent/libraries/cos_agent). However, if we click on [cos_agent](https://charmhub.io/integrations/cos_agent) from the interface table page, we can see that cos_agent is provded or required by multiple charms each have their own library link
image

This raises some questions to me:
- why is 1 library being featured over others to be included in the interfaces table?
- Does including 1 library link in the interface table force people to only focus on that library/that charm and not have the possibility to explore the other charms the interface provides/required for?
- should we be sharing all the libraries if an interface has more than 1 library? (that solve one problem but cause a huge UI issue)

A suggestion I have for this would be drop the library link and allow the user to explore the individual interface page and choose which library they want to dive into. Alternatively, to make it clear that an interface has libraries, we keep the library link but the URL is to the individual interface page (e.g. https://charmhub.io/integrations/cos_agent) and allow users to go to the library they want to

There are also the 'Other Charms' listed in the [individual interface page](https://charmhub.io/integrations/cos_agent), but they don't have any libraries so I have no comments for that as its also probably out of scope for this (here I'm a bit out of my depth as to why those charms are referenced on an individual interface page when those charms don't have any libraries, an explanation would be helpful plz)
 
 
 I believe we should keep the documentation link as it links to all the different interface version documentations so we don't have to worry about referring to each interface version's documentation link.
 
 As for the 'versions' column, how confident are we that we don't need to have that column? can we replace the copy on contributing (which was mentioned to remove as its outdated) with some information about how these interfaces are backwards compatible? 
 
 Pagination can be dropped only if the content of the table can be loaded in a adequate amount of time.
 
 I'm not able to see @james-garner-canonical's demo as the service is temporarily unavailable so I can't comment further before seeing your demo
 
### Charm 'Integrations' tab 
 Again as I was going through feedback, the demo service is temporarily unavailable, so I couldn't compare @james-garner-canonical's mock up to the demo for the charm page integration tab
 
 Once the demo service is available, I will finish going through through the rest of the feedback (which I think sound reasonable but I'd like to do a comparison before commenting) Again thank you so much for all the feedback and I'm onboard with the overall suggestions everyone has made :)

james-garner-canonical:
Thanks @migueldivo. Your thoughts on the Charmhub landing page sound good to me.

---

Regarding library links, the current idea for interfaces is that there is one canonical charm library per interface. In future, this will be exactly `charmlibs.interfaces.<interface name>`. Currently, the majority of interfaces still use legacy, Charmhub-hosted libraries, which have the quirk of being hosted under a specific charm (sometimes not the charm you expect!).

You raise a great point about the 'see libraries' links on each 'featured charms' card. This had completely slipped under my radar. I think the 'see libraries' link should be dropped entirely here -- these cards are for charm users to find charms to integrate with, and these just link to the libraries that that charm happens to host on Charmhub -- not very relevant for users, and unlikely to be relevant for the interface. (One of the libraries hosted under one of these charms might be a library for this interface, but we want to point charm authors directly to the canonical library anyway.)

To answer your question about the 'featured' and 'other' charms: the charms shown on these pages are (I believe) dynamically populated by Charmhub based on which charms require/provide the interface. The 'featured' ones are those explicitly listed in the interface metadata in `charmlibs`. I believe the 'featured' feature was developed to incentivise charm authors to test their charms with the interfaces. We've moved away from this approach for now, so the 'featured' charms are a bit baseless. I wouldn't be opposed to dropping the 'featured' feature entirely, but I didn't suggest it because I suspect that in future we'll end up providing 'featured' charms based on deeper QA and integration testing work, at which point the source of truth could just be changed over.

---

Regarding interface versioning, we have [some developer docs](https://documentation.ubuntu.com/charmlibs/how-to/design-relation-interfaces/) describing the backwards compatibility requirements. Perhaps it would make sense to add a sentence to the top-of-page text like "When designing an interface, [follow these guidelines](https://documentation.ubuntu.com/charmlibs/how-to/design-relation-interfaces/) to ensure backwards compatibility."?

Note that later this cycle we'll probably have URLs like https://documentation.ubuntu.com/charmlibs/reference/interfaces/cos_agent/ go directly to the docs for the latest interface version. In fact we'll probably collapse down to only storing the docs for the latest version in the monorepo at all, with the documented version number serving mostly as a starting point for forensic archeology.

---

Based on this, I continue to lean towards dropping the 'Links' column from the table. The Charmhub interface page should link to the library for charm developers, but Juju/charm users don't need it. The interface docs themselves are even lower-level, primarily for charm library developers (though can be useful for charmers for debugging), so I'd surface this even later on the interface page. So if we do retain the column, I'd prefer it to just have the one link to the canonical charm interface library.

---

Sorry if it was unclear in my comments -- the `demo.haus` links are all to @M7mdisk's work from this PR. They seem to be up for me right now (though slow to load). My suggestions are all based on them (or for the `integrations/foo` pages, based on the (identical) live version of the site).

migueldivo:
Hi @james-garner-canonical, thanks for the feedback. @M7mdisk and I have gone through it together and here is what we've decided.

For the landing page, we will include the links under the category filtering section, we won't add it to the footer yet as it can be easily missed by users not scrolling all the way to the bottom of the page (we don't get much traction in the footer either), and it might be nice to reassess the footer in the future once we have more developer resource links to include in charmhub and figure out how to better optimise the footer links.

For the interface table page we're happy with dropping the 'version' column and replacing the contribution copy at the top of the page with information + link about backwards compatibility. As for the 'links' column, we're committed to adding these links as we currently do not have another effective place to include these links for the relevant users to find. If we drop these links we risk charm developers (and charm users who want to learn and understand more about charms/libraries/etc) from not discovering that these resources exist. To make the links clearer that they are 2 separate links, we'll include the 'external link icon' similar to @lukasSerelis' design.
image

Your suggestions for the integrations tab within a charm are all reasonable changes we're happy to make them as they make the page clearer and reduces the duplication of text :)

Let us know if you have any further questions/feedback/comments

james-garner-canonical:
Thanks @migueldivo, the plan for the landing page, interface table page, and interface tab on charm pages sounds great.

> As for the 'links' column, we're committed to adding these links as **we currently do not have another effective place to include these links for the relevant users to find**. If we drop these links we risk charm developers (and charm users who want to learn and understand more about charms/libraries/etc) from not discovering that these resources exist.

I'd really like to see these links surfaced on the `interfaces/foo` pages, as suggested in [this comment](https://github.com/canonical/charmhub.io/pull/2356#issuecomment-4726355943), with explanatory text indicating the intended audience -- the library is for charm developers (irrelevant for Juju users), the low-level interface specification is primarily for charm library authors (largely irrelevant for charm developers). (Though this could be a separate PR.)

My remaining hesitation with the links column is twofold:
1. The 'external link icon' is helpful (though note that many 'Library' links will be to Charmhub, so not external), but I worry that it's not obvious from the link text where these links go, particularly 'Documentation' -- is it the library's documentation, or the low-level interface docs? 
2. These links essentially provide shortcuts deeper into the information heirarchy of the interface, which compete with the entrypoint we'd prefer people to go through -- the interface page, which should explain the nature of each link and its intended audience.

migueldivo:
Okay I understand now, the links are too low-level to be featured as primary entrypoints into an interface so we don't have to immediately surface them, so it would be better to include them within the `interface/foo` page and let users explore from there (depending on the type of user they are they will go to the appropriate developer resource)

As you mentioned, this is getting out of scope for this work, but I do suggest we follow a similar pattern as in the landing page of have these links under a general 'DEVELOPER RESOURCES' section on the LHS of the `interface/foo`, and we remove the 'links' column from the interface table page. If for now this works, please feel free to suggest other names for the library and documentation links

quick example done through page inspect
image




For the interface table page, could you give a quick review on this updated copy:

> Interfaces describe the relation between two charms. This interface catalogue shows opinionated, standardized interface specifications for charm relations, outlining the exact behavior and requirements of how charms interact with one another.
> 
> To maintain stability across the ecosystem, all interfaces adhere to a strict standard of backwards compatibility, ensuring seamless integration regardless of the interface version. When designing an interface, follow [these guidelines](https://documentation.ubuntu.com/charmlibs/how-to/design-relation-interfaces/) to ensure backwards compatibility.

james-garner-canonical:
Sorry for the delayed reply. 

---

> For the interface table page, could you give a quick review on this updated copy:
> 
> > Interfaces describe the relation between two charms. This interface catalogue shows opinionated, standardized interface specifications for charm relations, outlining the exact behavior and requirements of how charms interact with one another.
> > To maintain stability across the ecosystem, all interfaces adhere to a strict standard of backwards compatibility, ensuring seamless integration regardless of the interface version. When designing an interface, follow [these guidelines](https://documentation.ubuntu.com/charmlibs/how-to/design-relation-interfaces/) to ensure backwards compatibility.

This looks good. I'd drop the part about interface versions though:

> To maintain stability across the ecosystem, all interfaces adhere to a strict standard of backwards compatibility. When designing an interface, follow [these guidelines](https://documentation.ubuntu.com/charmlibs/how-to/design-relation-interfaces/) to ensure backwards compatibility.

---

> Okay I understand now, the links are too low-level to be featured as primary entrypoints into an interface so we don't have to immediately surface them, so it would be better to include them within the `interface/foo` page and let users explore from there (depending on the type of user they are they will go to the appropriate developer resource)
> 
> As you mentioned, this is getting out of scope for this work, but I do suggest we follow a similar pattern as in the landing page of have these links under a general 'DEVELOPER RESOURCES' section on the LHS of the `interface/foo`, and we remove the 'links' column from the interface table page. If for now this works, please feel free to suggest other names for the library and documentation links.

Including links in the sidebar sounds reasonable to me. I'd drop the "Specification archive" link (we would route there from lower-level content). So the two links would be (1) the canonical library for the interface, and (2) the low-level/wire interface specification. The concise names could be "Interface Library" and "Interface Documentation".

---

Regarding the `interface/foo` page, I'd love to have the short interface summary sentence shown in the table repeated prominently on the page.

I think it would also be really nice to include the longer interface description that [we ship in the `index.json`](https://github.com/canonical/charmlibs/blob/main/interfaces/index.json) on the page too.

If we do, then I also quite like the idea of adding a few automatically generated sentences based on what links are available for this interface, to explain the nature and audience of the links. Something like this mockup (ignore the side bar):

Screenshot From 2026-06-28 14-23-32

I'd propose generating the sentences like this
  - If there are any charms (probably always if the page exists at all?) "The charms that can be related over this interface are listed below." -- a sentence for charm users.
  - If the `lib` field is populated, add a sentence for charm authors: "To use this interface in your charm, use the <lib name> library." if the `lib_url` field is populated, make <lib name> a link.
    - If the (new) `lib_docs_url` field is present: "See the library docs for more.", with 'library docs' linking to `lib_docs_url`.
  - Finally if the `docs_url` field is present: "To learn more about how this interface works at a low level, read the interface docs.", with 'interface docs' linking to `docs_url`.

Screenshot From 2026-06-28 14-23-32-edit

This could totally be a separate, future PR, but I think having an idea of what would be shown on the `interfaces/foo` pages can help inform the higher-level content like the interfaces table.

edisile (approved):

PR #2359, opened 2026-06-12
Wd 37112/add solution local storage

abbiesims:
## Done
- Refreshes CSRF token  before submit to avoid stale token errors when publishers eventually submit metadata
- Adds local storage on edit solution page

## How to QA
- Go to https://charmhub-io-2359.demos.haus/solutions
- Edit a published solution and see the auto save message
- Refresh the page / go back a page and click edit again and you should see the same edits that you just made
- Test complex fields (useful links, maintainers, charms) and make sure changes are saved
- Click cancel after edits and make sure you see the modal asking if you're sure you want to cancel and lose local changes
- Submit the form and make sure it submits correctly

## Issue / Card

Fixes [WD-37112](https://warthogs.atlassian.net/browse/WD-37112)

[WD-37112]: https://warthogs.atlassian.net/browse/WD-37112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/src/solutions/modules/ConfirmationModal.js
+++ b/static/js/src/solutions/modules/ConfirmationModal.js
@@ -15,6 +15,10 @@ function initConfirmationModal(config) {
   let submitter = null;
   const originalButtonText = confirmButton.textContent;
 
+  const hideSubmitError = () => {
+    document.getElementById("solution-submit-error")?.classList.add("u-hide");
+  };
+
   const showSubmitError = (message) => {
     const notification = document.getElementById("solution-submit-error");
     notification.querySelector(".p-notification__message").textContent =
@@ -33,6 +37,7 @@ function initConfirmationModal(config) {
         formSubmitted = false;
         confirmButton.textContent = originalButtonText;
         confirmButton.disabled = false;
+        modal.classList.add("u-hide");
         showSubmitError(
           error.message || "Unable to submit the form. Please try again."
         );
@@ -53,6 +58,7 @@ function initConfirmationModal(config) {
     }
 
     submitter = event.submitter;
+    hideSubmitError();
 
     event.preventDefault();
     event.stopPropagation();
--- a/static/js/src/solutions/modules/ConfirmationModal.js
+++ b/static/js/src/solutions/modules/ConfirmationModal.js
@@ -13,9 +13,22 @@ function initConfirmationModal(config) {
 
   let formSubmitted = false;
   let submitter = null;
+  const originalButtonText = confirmButton.textContent;
 
-  const submitForm = () => {
+  const submitForm = async () => {
     formSubmitted = true;
+
+    if (typeof config.beforeSubmit === "function") {
+      try {
+        await config.beforeSubmit(form, submitter);
+      } catch (error) {
+        formSubmitted = false;
+        confirmButton.textContent = originalButtonText;
+        confirmButton.disabled = false;
+        alert(error.message || "Unable to submit the form. Please try again.");
+        return;
+      }
+    }
 
     if (form.requestSubmit) {
       form.requestSubmit(submitter || undefined);
@@ -56,8 +69,6 @@ function initConfirmationModal(config) {
     modal.classList.remove("u-hide");
     confirmButton.focus();
   });
-
-  const originalButtonText = confirmButton.textContent;
 
   confirmButton.addEventListener("click", () => {
     const spinner = document.createElement("i");
--- a/static/js/src/solutions/modules/ConfirmationModal.js
+++ b/static/js/src/solutions/modules/ConfirmationModal.js
@@ -21,8 +21,16 @@ function initConfirmationModal(config) {
 
   const showSubmitError = (message) => {
     const notification = document.getElementById("solution-submit-error");
-    notification.querySelector(".p-notification__message").textContent =
-      message;
+    const notificationMessage = notification?.querySelector(
+      ".p-notification__message"
+    );
+
+    if (!notification || !notificationMessage) {
+      console.error("Submit error notification: element not found");
+      return;
+    }
+
+    notificationMessage.textContent = message;
     notification.classList.remove("u-hide");
     notification.scrollIntoView({ behavior: "smooth", block: "center" });
   };
--- a/static/js/src/solutions/modules/ConfirmationModal.js
+++ b/static/js/src/solutions/modules/ConfirmationModal.js
@@ -15,6 +15,14 @@ function initConfirmationModal(config) {
   let submitter = null;
   const originalButtonText = confirmButton.textContent;
 
+  const showSubmitError = (message) => {
+    const notification = document.getElementById("solution-submit-error");
+    notification.querySelector(".p-notification__message").textContent =
+      message;
+    notification.classList.remove("u-hide");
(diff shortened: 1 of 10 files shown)


static/js/src/solutions/modules/ConfirmationModal.js
```diff
@@ -13,10 +13,23 @@ function initConfirmationModal(config) {
...
     formSubmitted = true;
 
+    if (typeof config.beforeSubmit === "function") {
+      try {
+        await config.beforeSubmit(form, submitter);
+      } catch (error) {
+        formSubmitted = false;
+        confirmButton.textContent = originalButtonText;
+        confirmButton.disabled = false;
+        alert(error.message || "Unable to submit the form. Please try again.");
```
bartaz: Using browser `alert` seems a bit barebones. Is there any existing form validation, or a notification we could use here?

templates/solutions/edit-solution.html
```diff
@@ -82,6 +82,9 @@ <h1 class="p-heading--2">
           </button>
         {% endif %}
       </div>
+      <p id="local-autosave-status" class="p-form-help-text u-no-margin--bottom u-no-max-width" aria-live="polite">
```
bartaz: This feels a bit cramped with the "* required fields" text below the line. Maybe the autosave would work better on right hand below buttons?

Image

templates/solutions/edit-solution.html
```diff
@@ -115,5 +118,20 @@ <h1 class="p-heading--2">
...
 
 {% include 'solutions/partials/_confirmation_modal.html' %}
+<div class="p-modal u-hide" id="local-autosave-cancel-modal">
+  <section class="p-modal__dialog" role="dialog" aria-modal="true" aria-labelledby="local-autosave-cancel-title" aria-describedby="local-autosave-cancel-description">
+    <header class="p-modal__header">
+      <h2 class="p-modal__title" id="local-autosave-cancel-title">Discard local changes?</h2>
+      <button type="button" class="p-modal__close" aria-label="Close modal" aria-controls="local-autosave-cancel-modal">Close</button>
+    </header>
+    <div class="p-modal__body" id="local-autosave-cancel-description">
+      <p>Your browser has autosaved edits for this solution. If you cancel, these local edits will be removed. Changes already submitted to Charmhub will not be affected.</p>
```
bartaz: The text "if you cancel, these local edits will be removed" is a bit confusing, as there is no cancel button on the modal. Maybe "if you leave" would work better in context of "Discard and leave" button?

static/js/src/solutions/modules/LocalAutosave.js
```diff
@@ -0,0 +1,354 @@
+const DRAFT_STORAGE_PREFIX = "solution-edit-draft";
+const DRAFT_SAVE_DELAY = 500;
+const DRAFT_MAX_AGE = 7 * 24 * 60 * 60 * 1000;
+const REPEATED_FIELD_CONFIGS = [
```
bartaz: With the list of the fields hardcoded here, does it mean that if the edit form changes this would need to be updated as well?

bartaz (commented):
Looks fine, couple questions and comments inline in code.

static/js/src/solutions/modules/LocalAutosave.js
abbiesims: Yes, but only for the "complex" field. I can add some `data` attributes to the templates instead maybe? Let me know what you think

bartaz:
After confirming submit if error happens modal is not closed, so the error notification is not visible.

image

static/js/src/solutions/modules/ConfirmationModal.js
```diff
@@ -13,10 +13,33 @@ function initConfirmationModal(config) {
 
   let formSubmitted = false;
   let submitter = null;
+  const originalButtonText = confirmButton.textContent;
+
+  const showSubmitError = (message) => {
+    const notification = document.getElementById("solution-submit-error");
+    notification.querySelector(".p-notification__message").textContent =
+      message;
+    notification.classList.remove("u-hide");
```
bartaz, line 34: Do we ever hide it again? Or do we depend on the page being reloaded for it do be removed?

static/js/src/solutions/modules/LocalAutosave.js
```diff
@@ -0,0 +1,326 @@
+const DRAFT_STORAGE_PREFIX = "solution-edit-draft";
+const DRAFT_SAVE_DELAY = 500;
+const DRAFT_MAX_AGE = 7 * 24 * 60 * 60 * 1000;
```
bartaz, line 3: Do we need the max age? Is 1 week enough?

static/js/src/solutions/modules/LocalAutosave.js
```diff
@@ -0,0 +1,326 @@
...
+
+    if (!skipBeforeUnloadSave && hasChanges()) {
+      saveTimeout = window.setTimeout(saveChanges, DRAFT_SAVE_DELAY);
+    }
+  };
+
+  initCancelLink(form, discardDraft);
+  form.addEventListener("input", scheduleSave);
+  form.addEventListener("change", scheduleSave);
+  form.addEventListener("click", () => window.setTimeout(scheduleSave, 0));
```
bartaz, line 321: if there is a reason to wrap it in `setTimeout` while others are not, maybe good to add a comment to explain why it's treated differently

static/js/src/solutions/modules/LocalAutosave.js
abbiesims: I've put it so that drafts in local storage don't sit there forever. I'll be adding a proper "save draft" functionality that won't have en expiry, the local storage solution is mainly if there happen to be any authentication errors or the user closes the tab by accident. Happy to increase the max age if you think it's necessary?

static/js/src/solutions/modules/ConfirmationModal.js
```diff
@@ -13,10 +13,46 @@ function initConfirmationModal(config) {
...
+    document.getElementById("solution-submit-error")?.classList.add("u-hide");
+  };
+
+  const showSubmitError = (message) => {
+    const notification = document.getElementById("solution-submit-error");
+    const notificationMessage = notification?.querySelector(
+      ".p-notification__message"
+    );
+
+    if (!notification || !notificationMessage) {
```
bartaz, line 28: It doesn't hurt, but I'm not sure how useful it is. 
If the notification HTML is not in place, probably something bigger is wrong, so showing an alert instead may not be exactly helpful. At this point it would likely be fair to just log error to the console?

But overall, I don't mind it staying.

static/js/src/solutions/modules/LocalAutosave.js
bartaz: We can adjust it later if needed. It will be less relevant if the proper save draft will be in place.

bartaz (approved):
LGTM, thanks.

static/js/src/solutions/modules/ConfirmationModal.js
abbiesims: Sure, made the change actually as I think you're right

PR #2409, opened 2026-08-27
chore: workshop setup

M7mdisk:
## Done
- setup workshop
- remove cypress check (its always failing)
## How to QA
- pretty much the same as whats in https://github.com/canonical/snapcraft.io/pull/5735

### DIFF
--- a/.workshop/charmhub-io/env.sh
+++ b/.workshop/charmhub-io/env.sh
@@ -1,0 +1,42 @@
+# shellcheck shell=bash
+# Shared environment for the `charmhub-io` workshop actions and SDK hooks.
+#
+# dotrun used to export `.env`/`.env.local` and put the project venv on PATH.
+# Nothing does that automatically here, so actions source this file instead.
+
+PROJECT_DIR=/project
+cd "$PROJECT_DIR"
+
+set -a
+if [ -f "$PROJECT_DIR/.env" ]; then
+  . "$PROJECT_DIR/.env"
+fi
+if [ -f "$PROJECT_DIR/.env.local" ]; then
+  . "$PROJECT_DIR/.env.local"
+fi
+set +a
+
+# Sourced by the actions AND by ~/.profile, so guard against PATH entries
+# piling up in nested shells.
+for _charmhub_dir in "$PROJECT_DIR/.venv/bin" "$PROJECT_DIR/node_modules/.bin"; do
+  case ":$PATH:" in
+    *":$_charmhub_dir:"*) ;;
+    *) PATH="$_charmhub_dir:$PATH" ;;
+  esac
+done
+unset _charmhub_dir
+export PATH
+
+# The node SDK ships Corepack; fall back to it when no yarn shim is on PATH.
+# This has to be a real executable rather than a shell function: `yarn start`
+# runs concurrently, which spawns `sh -c 'yarn run serve'`, and an exported
+# bash function would not survive into that subshell.
+if ! command -v yarn >/dev/null 2>&1; then
+  CHARMHUB_IO_SHIM_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/charmhub-io-workshop/bin"
+  mkdir -p "$CHARMHUB_IO_SHIM_DIR"
+  if [ ! -x "$CHARMHUB_IO_SHIM_DIR/yarn" ]; then
+    printf '#!/bin/bash\nexec corepack yarn "$@"\n' > "$CHARMHUB_IO_SHIM_DIR/yarn"
+    chmod +x "$CHARMHUB_IO_SHIM_DIR/yarn"
+  fi
+  export PATH="$CHARMHUB_IO_SHIM_DIR:$PATH"
+fi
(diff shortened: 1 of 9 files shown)


.workshop/charmhub-io/hooks/setup-base
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# setup-base: runs as root, before the project is mounted.
+
+set -euo pipefail
+
+# Native build deps for the Python/Node dependency trees.
+# Evidence: Dockerfile python-dependencies stage.
+apt-get update
+apt-get install --no-install-recommends --yes \
```
edisile, line 9: I don't think we actually need these, they're not even real dependencies for the project... on snapcraft this step is empty and it works fine.

Also, why these are in the Dockerfile is a mystery to me

.workshop/charmhub-io/hooks/check-health
edisile, line 1: cool! I didn't know about this hook

workshop.yaml
```diff
@@ -0,0 +1,70 @@
...
+    ENVIRONMENT=prod yarn serve
+
+  install: |
+    bash /project/.workshop/charmhub-io/install.sh
+
+  test: |
+    source /project/.workshop/charmhub-io/env.sh
+    yarn test
+
+  # One intent, three linters.
```
edisile: absolute slop comment lmao

workshop.yaml
```diff
@@ -0,0 +1,70 @@
...
+      vite:
+        interface: tunnel
+        endpoint: localhost:5045 # vite dev server
+
+# Only intents that add logic get an action; plain package.json scripts go
+# through the `yarn` escape hatch, so script names live in one place.
+actions:
+  # Equivalent of the old `dotrun`: Vite + gunicorn together on 8045/5045.
+  start: |
+    source /project/.workshop/charmhub-io/env.sh
```
edisile: there's no need to explicitly source the script here (or any of the following actions), it's sourced through `~/.profile`

workshop.yaml
```diff
@@ -0,0 +1,70 @@
...
+  build: |
+    source /project/.workshop/charmhub-io/env.sh
+    yarn build
+
+  prod: |
+    source /project/.workshop/charmhub-io/env.sh
+    yarn build
+    ENVIRONMENT=prod yarn serve
+
+  install: |
```
edisile, line 43: good point, I forgot dotrun had `clean` and `install` commands, I should add them to the snapcraft workshop as well

.workshop/charmhub-io/install.sh
edisile, line 1: lots of noise in this file... it can be cleaned up a lot:
- this script always runs in `/project` as the workshop user both during `setup-project` (confirmed by [docs](https://ubuntu.com/workshop/docs/reference/sdks/#sdk-hooks:~:text=The%20exception%20is%20setup%2Dproject%2C%20which%20runs%20as%20the%20workshop%20user%20in%20the%20%2Fproject%2F%20directory)) and when called as an action, so all the explicit paths can be removed
- no need for that `! -x` guard on the venv python, just `uv venv --allow-existing`
- sourcing the env does nothing in this step (unless we're doing stuff like override node, uv or python env vars and I don't think we are)

workshop.yaml
```diff
@@ -0,0 +1,76 @@
...
+    yarn start
+
+  build: |
+    source /project/.workshop/charmhub-io/env.sh
+    yarn build
+
+  prod: |
+    source /project/.workshop/charmhub-io/env.sh
+    yarn build
+    ENVIRONMENT=prod yarn serve
```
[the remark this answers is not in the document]
edisile: we don't need any of this

.workshop/charmhub-io/env.sh
```diff
@@ -0,0 +1,42 @@
...
+# bash function would not survive into that subshell.
+if ! command -v yarn >/dev/null 2>&1; then
+  CHARMHUB_IO_SHIM_DIR="${XDG_CACHE_HOME:-$HOME/.cache}/charmhub-io-workshop/bin"
+  mkdir -p "$CHARMHUB_IO_SHIM_DIR"
+  if [ ! -x "$CHARMHUB_IO_SHIM_DIR/yarn" ]; then
+    printf '#!/bin/bash\nexec corepack yarn "$@"\n' > "$CHARMHUB_IO_SHIM_DIR/yarn"
+    chmod +x "$CHARMHUB_IO_SHIM_DIR/yarn"
+  fi
+  export PATH="$CHARMHUB_IO_SHIM_DIR:$PATH"
+fi
```
edisile, line 42: why would the `yarn` shim not be in PATH?

.workshop/charmhub-io/env.sh
```diff
@@ -0,0 +1,42 @@
...
+  . "$PROJECT_DIR/.env"
+fi
+if [ -f "$PROJECT_DIR/.env.local" ]; then
+  . "$PROJECT_DIR/.env.local"
+fi
+set +a
+
+# Sourced by the actions AND by ~/.profile, so guard against PATH entries
+# piling up in nested shells.
+for _charmhub_dir in "$PROJECT_DIR/.venv/bin" "$PROJECT_DIR/node_modules/.bin"; do
```
edisile, line 21: why not activate the .venv normally? also I thought adding `node_modules/.bin` to PATH was bad form?

